PR #12, opened 2022-01-10
Remove unnecessary println

jphastings:
This seems to have been left in from a long time ago!


mochi-co:
Wow that is super old! Nice catch!

mochi-co:
@jphastings This fix has been merged and released as `v1.0.2`! Thanks for your help!

jphastings:
My pleasure — thanks!

PR #18, opened 2022-01-24
OnConnect and OnDisconnect Event Hooks

mochi-co:
Adds OnConnect and OnDisconnect Event hooks, allowing embedding services to gain information about the connection and disconnection of clients.


mochi-co:
- All unit tests passing
- Paho interoperability tests passing

Issue #17, opened 2022-01-23
SIGSEGV crash when running examples/tcp/main.go


ogelami:
I noticed this issue when i switched from my PC to my raspberry pi, trying to narrow down the cause it seems like i get the same SIGSEGV error when running the tcp example.

Running on Raspberry Pi 3 Model B Plus Rev 1.3, ARMv7 Processor rev 4 (v7l). I've confirmed that there are no issues to bind to the port.

```
root@housekeeper:/home/ogelami/mochi-mqtt-issue# go run main.go

Mochi MQTT Server initializing... TCP
  Started!
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x1238c]

goroutine 22 [running]:
runtime/internal/atomic.goLoad64(0x1cae03c, 0x0, 0x0)
        /usr/lib/go-1.15/src/runtime/internal/atomic/atomic_arm.go:131 +0x1c
github.com/mochi-co/mqtt/server/listeners.(*TCP).Serve(0x1cae000, 0x1c9a028)
        /root/go/pkg/mod/github.com/mochi-co/mqtt@v1.0.4/server/listeners/tcp.go:89 +0x28
github.com/mochi-co/mqtt/server/listeners.(*Listeners).Serve.func1(0x1cac000, 0x29db60, 0x1cae000, 0x1c9a028)
        /root/go/pkg/mod/github.com/mochi-co/mqtt@v1.0.4/server/listeners/listeners.go:94 +0x70
created by github.com/mochi-co/mqtt/server/listeners.(*Listeners).Serve
        /root/go/pkg/mod/github.com/mochi-co/mqtt@v1.0.4/server/listeners/listeners.go:91 +0x9c
exit status 2
```

mochi-co:
Interesting. If I had to guess it's because ARMv7 is a 32bit chip, whereas the atomic counters in the broker are set to use 64bit ints. Is this something you are relying on? I can have a think about how this issue can be fixed, and I welcome suggestions from anyone else!

ogelami:
Thanks for the quick response! Shouldn't 32-bit be enough for the broker counters?

Could using `int` where possible and wrapping the `sync/atomic` with 32/64-bit check for calling say AddUint64 or AddUint32 be a partial solution?

```
package atomthic

import (
	"sync/atomic"
)

const uintSize = 32 << (^uint(0) >> 32 & 1)

[...]

PR #19, opened 2022-01-25
Improve 32-bit compatibility

rkennedy:
This is mainly a fix for #17, where 64-bit variables may be misaligned for atomic operations on 32-bit systems. I think I found all struct fields that are used in atomic operations. Those that are used as simple flags are changed to `uint32` so they'll be aligned correctly on all systems. Some fields are used for counting things, so I assumed they might really need more than 32 bits. For those, I added tests that will check that alignment requirements are meet even when the tests run on 64-bit systems.

While investigating the atomic operations, I noticed a race condition between atomic loads and stores, and fixed it with a compare-and-swap operation.

Finally, the test for `encodeLength` assumed that `int` and `int64` always have the same range. I had to change the function signature to get the tests to pass.

### DIFF
--- a/server/internal/circ/buffer.go
+++ b/server/internal/circ/buffer.go
@@ -24,12 +24,13 @@ type Buffer struct {
 	block int          // the size of the R/W block.
 	buf   []byte       // the bytes buffer.
 	tmp   []byte       // a temporary buffer.
+	_     int32        // align the next fields to an 8-byte boundary for atomic access.
 	head  int64        // the current position in the sequence - a forever increasing index.
 	tail  int64        // the committed position in the sequence - a forever increasing index.
 	rcond *sync.Cond   // the sync condition for the buffer reader.
 	wcond *sync.Cond   // the sync condition for the buffer writer.
-	done  int64        // indicates that the buffer is closed.
-	State int64        // indicates whether the buffer is reading from (1) or writing to (2).
+	done  uint32       // indicates that the buffer is closed.
+	State uint32       // indicates whether the buffer is reading from (1) or writing to (2).
 }
 
 // NewBuffer returns a new instance of buffer. You should call NewReader or
@@ -127,7 +128,7 @@ func (b *Buffer) awaitEmpty(n int) error {
 	// then wait until tail has moved.
 	b.rcond.L.Lock()
 	for !b.checkEmpty(n) {
-		if atomic.LoadInt64(&b.done) == 1 {
+		if atomic.LoadUint32(&b.done) == 1 {
 			b.rcond.L.Unlock()
 			return io.EOF
 		}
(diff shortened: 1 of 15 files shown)


ogelami:
Tested and i can confirm that this now runs on a 32-bit ARMv7

mochi-co:
@ogelami Great news - nice work @rkennedy! 

I will pull this down as soon as I have a bit of time and run the usual tests / performance benchmarking to make sure everything checks out, otherwise it looks promising. Given the significant changes here, I am thinking this will form the basis for a v1.1.0 release.

mochi-co:
I've merged this PR into a new branch - mochi-co:feature/32bit-compatibility - to properly encompass the changes we want to make. I also dusted off the old pi zero to give it a spin so we're all testing in the same 32bit ARM environment.

My general thoughts:
- I like the changes of the state indicators from int64 to uint32 (done, state, end, etc). I think this makes a lot of sense as there's really no reason for them to be 64bit, regardless of the build architecture.
- I also particularly liked the implementation of CompareAndSwapUint32, that's a much more elegant solution.
- By default (as in, currently hard-coded), the buffer block size is 256*1024, as this was the range I found to be optimal for performance in the vast majority of use cases, so uint32 is more than sufficient to hold this range. If someone needs a buffer size greater than uint32, that will need to be a separate issue. That I can see, I don't believe any of these uint32 changes should negatively impact the performance for 64bit versions.

As a note - I ran some of my own experiments last night, and I was able to build and run the server on ARM32 using _only_ 8bit field alignment changes throughout the codebase. While we can probably resolve this issue altogether by optimising the fields and allowing the compiler to solve the problem for us (as per https://pkg.go.dev/sync/atomic#pkg-notes), I'm really happy with the above changes and I'm going to merge them into the new 32bit branch as-is. Once I've done this, I will finish my optimisations to ensure all fields are 8bit aligned, then we can give it a final review and make sure everything works as intended, and, we'll call it v1.1.0.

mochi-co (commented):
I've merged this PR into a new branch - mochi-co:feature/32bit-compatibility - to properly encompass the changes we want to make. I also dusted off the old pi zero to give it a spin so we're all testing in the same 32bit ARM environment.

My general thoughts:

I like the changes of the state indicators from int64 to uint32 (done, state, end, etc). I think this makes a lot of sense as there's really no reason for them to be 64bit, regardless of the build architecture.
I also particularly liked the implementation of CompareAndSwapUint32, that's a much more elegant solution.
By default (as in, currently hard-coded), the buffer block size is 256*1024, as this was the range I found to be optimal for performance in the vast majority of use cases, so uint32 is more than sufficient to hold this range. If someone needs a buffer size greater than uint32, that will need to be a separate issue. That I can see, I don't believe any of these uint32 changes should negatively impact the performance for 64bit versions.

As a note - I ran some of my own experiments last night, and I was able to build and run the server on ARM32 using only 8bit field alignment changes throughout the codebase. While we can probably resolve this issue altogether by optimising the fields and allowing the compiler to solve the problem for us (as per https://pkg.go.dev/sync/atomic#pkg-notes), I'm really happy with the above changes and I'm going to merge them into the new 32bit branch as-is. Once I've done this, I will finish my optimisations to ensure all fields are 8bit aligned, then we can give it a final review and make sure everything works as intended, and, we'll call it v1.1.0.

PR #26, opened 2022-02-23
added redis persistence mode

wind-c:
1. Added redis persistence mode
2. Added redis persistence example
3. Fixed cannot replace old retained message bug
4. Fixed type assignment bug when persistent inflight message

### DIFF
--- a/examples/redis/main.go
+++ b/examples/redis/main.go
@@ -1,0 +1,60 @@
+package main
+
+import (
+	"fmt"
+	"github.com/go-redis/redis/v8"
+	"github.com/logrusorgru/aurora"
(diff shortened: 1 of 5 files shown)


mochi-co:
@wind-c These look like very well thought out and I'm certain a lot of people will benefit from these changes! They pass the usual paho/unit tests, so I'm going to merge them as `v1.1.2` 

Thank you!

mochi-co:
@wind-c I just noticed that the redis store tests fail unless redis is running, so we'll need to ensure the tests can be run without the third-party server, otherwise the build will fail forever. I've reverted the changes for now but it seems I can't reopen this request. Would you be able to re-open the pull request so the test changes can be made?

PR #28, opened 2022-02-25
added redis persistence mode

wind-c:
1. Added redis persistence mode
2. Added redis persistence example
3. Fixed cannot replace old retained message bug
4. Fixed type assignment bug when persistent inflight message

### DIFF
--- a/server/persistence/redis/redis.go
+++ b/server/persistence/redis/redis.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/json"
 	"errors"
+	"fmt"
 	"github.com/go-redis/redis/v8"
 	"github.com/mochi-co/mqtt/server/persistence"
 	"github.com/mochi-co/mqtt/server/system"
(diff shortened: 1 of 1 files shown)


mochi-co:
Hi @wind-c - not sure if something went wrong with the PR, as it seems like most of the code from the original is missing. I think what we really want to do it use the code from the original PR and merge it into the redis branch (`origin/persist-redis`). Then we need to use a mock redis handler in the tests instead of connecting to a real redis server (you or someone else can do this once it's merged to the redis branch). This should make it releasable.

wind-c:
> Hi @wind-c - not sure if something went wrong with the PR, as it seems like most of the code from the original is missing. I think what we really want to do it use the exact code from the original PR and merge it into the redis branch (`origin/persist-redis`). Then we need to use a mock redis handler in the tests instead of connecting to a real redis server (you or someone else can do this once it's merged to the redis branch). This should make it releasable.

Ok，I will check and resubmit complete PR.

PR #29, opened 2022-02-27
Support non-UTF8 payloads (per MQTT specification)

jmacd:
**What are you trying to achieve?**

I'm testing this MQTT broker service with the use of Sparkplug-B payloads. The payloads are Google protobuf encoded, are not and should not be expected to be valid UTF8. The is-valid-UTF8 test was causing my device's connection to fail during setup because the Will Message was invalid protobuf data.

The MQTT specification does not require UTF8 for []byte payloads, only for certain string fields, as I understand it.

Note: While testing, I discovered that ErrOffsetBytesOutOfRange was not being used. The `decodeBytes()` method was returning the error for string decoding. I have updated/improved the tests of the decode methods to test for expected errors explicitly and added an invalid-UTF8 test case.

**Why are you doing this?**

Correcting the use of errors in `codec.go` won't actually help propagate the reason for the failure back to my device. In `packets.go`, there is a block of code that discards the lower-level error:

```
		pk.WillMessage, offset, err = decodeBytes(buf, offset)
		if err != nil {
			return ErrMalformedWillMessage
		}
```

I'm not sure what the authors wishes are here -- discarding the low-level error is certainly efficient but makes diagnostics more difficult. I would prefer to see

```
		pk.WillMessage, offset, err = decodeBytes(buf, offset)
		if err != nil {
			return fmt.Errorf("%s: %w", err, ErrMalformedWillMessage)
		}
```

This returns the high-level error type information to the caller while also returning the low-level error explanation to the user. The cost, of course, is more allocation. Since errors should be unexpected, I prefer to think of this as not a performance hit. 

If you approve, I would be glad to apply the above suggestion throughout `packets.go` in a future PR.

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,7 +23,7 @@ MQTT stands for MQ Telemetry Transport. It is a publish/subscribe, extremely sim
 - Ring Buffer packet codec.
 - TCP, Websocket, (including SSL/TLS) and Dashboard listeners.
 - Interfaces for Client Authentication and Topic access control.
-- Redis and Bolt persistence and storage interfaces (see examples folder).
+- Bolt persistence and storage interfaces (see examples folder).
 - Directly Publishing from embedding service (`s.Publish(topic, message, retain)`).
 - Basic Event Hooks (currently `OnMessage`, `OnConnect`, `OnDisconnect`).
 - ARM32 Compatible.
--- a/server/internal/packets/codec.go
+++ b/server/internal/packets/codec.go
@@ -28,6 +28,10 @@ func decodeString(buf []byte, offset int) (string, int, error) {
 		return "", 0, err
 	}
 
+	if !validUTF8(b) {
+		return "", 0, ErrOffsetStrInvalidUTF8
+	}
+
 	return bytesToString(b), n, nil
 }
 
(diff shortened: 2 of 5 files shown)


jmacd:
Additional cleanups: I've removed the `result` field from test expectations where `shouldFail` is non-nil, since the unused `result` field in those cases is misleading. The `TestDecodeString` method's `result` type changes, only the first element of the slice was being used.

mochi-co:
> I'm not sure what the authors wishes are here 
That would be me! The answer is twofold - I was simply trying to be consistent in returns, and because it never really came up. 

Now that it has I am grateful for your contribution! I think this is an excellent idea and I'd like to include it in v1.1.2, as well as any relevant changes to `packets.go`. Big fan of `%w`, too.

I'll merge this to the `v1.1.2` branch 👍🏻

jmacd:
I'm still working on some diagnostics, maybe for a future release?

I found at least one lost error that was responsible for making it unclear
why a disconnect was happening. But on further investigation, the
disconnect was voluntary by the client and nothing wrong was happening in
the broker. So, not a high priority. I'll follow up.

For what it's worth, OpenTelemetry (my main project at work) has a bunch of
issues surrounding how to model tracing through a messaging system, e.g.,
see the discussion on
https://github.com/open-telemetry/opentelemetry-specification/pull/2278.
I've begun thinking that this exercise to instrument mochi-co/mqtt will
help me understand the problem, so you can expect me to send you
diagnostics improvements in the future.

PR #31, opened 2022-02-28
Add redis persistence mode and example

wind-c:
With redis as the datastore, services can be separated from storage, services can be redeployed to any machine, and data can be recovered。

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,7 +23,7 @@ MQTT stands for MQ Telemetry Transport. It is a publish/subscribe, extremely sim
 - Ring Buffer packet codec.
 - TCP, Websocket, (including SSL/TLS) and Dashboard listeners.
 - Interfaces for Client Authentication and Topic access control.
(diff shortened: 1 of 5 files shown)


mochi-co:
https://github.com/alicebob/miniredis might be a good solution for implementing a mock redis connection for these tests

mochi-co:
Many thanks @wind-c, sorry for the delay I've been very busy. I will try to look at this this evening!

mochi-co:
Looks good @wind-c ! Sorry for the delay, I got caught up with life things.

There's a handful issues remaining with this PR - for example, if no results are returned from redis, then redis returns a `redis.Nil` error, which in turn causes the server to register a failure. I am going to merge it into the `persist-redis` branch and then patch it where necessary. After that, it would be great if you could checkout the final changes and make sure it works for your use-case. 

If successful, we can merge it into master.

Issue #25, opened 2022-01-30
Some tests fail due to test timing issues when using -count=100 / 1000


mochi-co:
As per @rkennedy's comments in #24, various tests fail intermittently and typically exhibit when running `go test ./... -count=100`. Backtesting indicates these failures occur at least as early as v1.0.5. The failures can occur as few as 1 in every 500 runs. This is likely to be a problem with the tests rather than a problem with the broker code.

Determine the precise cause for these intermittent failures and correct it.

After running `go test ./... -count=100` several times, the known failures are as follows: 

```
writing err io: read/write on closed pipe
--- FAIL: TestServerResendClientInflightBackoff (0.01s)
    server_test.go:2096: 
                Error Trace:    server_test.go:2096
                Error:          Not equal: 
                                expected: []byte{0x3a, 0xe, 0x0, 0x5, 0x61, 0x2f, 0x62, 0x2f, 0x63, 0x0, 0xb, 0x68, 0x65, 0x6c, 0x6c, 0x6f}
                                actual  : []byte{}
                            
                                Diff:
                                --- Expected
                                +++ Actual
                                @@ -1,3 +1,2 @@
                                -([]uint8) (len=16) {
                                - 00000000  3a 0e 00 05 61 2f 62 2f  63 00 0b 68 65 6c 6c 6f  |:...a/b/c..hello|
                                +([]uint8) {
                                 }
                Test:           TestServerResendClientInflightBackoff
```

```
writing err io: read/write on closed pipe
--- FAIL: TestServerResendClientInflightBackoff (0.00s)
    server_test.go:2105: 
                Error Trace:    server_test.go:2105
                Error:          Not equal: 
                                expected: 1
                                actual  : 2
                Test:           TestServerResendClientInflightBackoff
```

```
2022/01/30 10:16:33 error writing to buffer io.Writer; io: read/write on closed pipe
--- FAIL: TestServerWriteClient (0.01s)
    server_test.go:671: 
                Error Trace:    server_test.go:671
                Error:          Not equal: 
                                expected: []byte{0x70, 0x2, 0x0, 0xe}
                                actual  : []byte{}
                            
                                Diff:
                                --- Expected
                                +++ Actual
                                @@ -1,3 +1,2 @@
                                -([]uint8) (len=4) {
                                - 00000000  70 02 00 0e                                       |p...|
                                +([]uint8) {
                                 }
                Test:           TestServerWriteClient
```

```
[...]

PR #34, opened 2022-03-08
Add OnError handler option and report the error that causes disconnects

### LINKED ISSUE #21: client net.Conn in onConnect Hook
In onConnect event it would be interesting to be able to read client net.Conn. For example to view remoteAddr.

### LINKED ISSUE #33: Show the reason for client disconnects
While debugging an application, I was at some point confounded by what looked to both the client and the server like the other had side had closed the connection. On further investigation, there were two identical clients, one a zombie process, that were both trying to connect, and each time one connects the other is disconnected.

The errors that were printed in OnDisconnect both do not include the remote address field, which revealed the presence of duplicate client IDs, these also happened to look like ordinary client-initiated disconnects, so the server logs would not reveal there to be a session takeover.

Auditing the code, I learned that both an explicit Disconnect packet could trigger disconnection, but so could EOF or a variety of errors in several code paths. The existing Client was well synchronized for this, but the original cause of stopping the connection was being lost. This adds the first cause to stop the connection. In EstablishConnection(), the root cause will be the one passed to OnDisconnect, so the server logs will reveal the root of the problem. (Since this is MQTT 3.x, there is no way for the Disconnect packet to tell the client, so it could place an explanation in its own logs.)

In addition, there would be no way for a user to configure visibility into other errors, which might less obviously be the cause of or associated with disconnection or simply provide warning flags. Proposing to add an `OnError` handler to capture support logging all errors. 

Related to #21, since if there are duplicate clients it will be nice to say where they're coming from.

jmacd:
Fixes #21. 
Fixes #33.

This is the result of auditing error handling in order to reveal the reason for disconnects, particularly when two clients are trying to use the same ID. Adds an `OnError` handler to report errors that do not directly cause disconection.

Adds error wrapping with %w in cases where the error alone might not reveal the code path.

Introduces an interface to pass Client info into the `events` package. The inline connection used for direct pub/sub stubs this out, so that errors related to the inline connection show properly. This allows the error handlers to decide to print the RemoteAddr().String().

Editorial: don't need to heap-allocate `sync.WaitGroup` the way it's used here, removes a few lines of initialization.
Editorial: removes a few instances of `err := call(); if err != nil { return err } return nil`, replaced with `return call()`.
Editorial: renames `Client.Read()` to `Client.ReadLoop()` for readability.

Call out a few places where errors are being ignored on purpose using `_ =`; note that one such location appears to be accidental, the call to `ResendClientInflight()`? I've kept the status-quo, which is to skip past the error, but it will use the new `OnError` handler. A new error code `ErrConnectionFailed` is created to signal auth-failure (which didn't necessary generate a non-nil error, only a non-success ReturnCode for the response).

Distinguish the two code paths that stop clients: StopWithClientLocked() is called during session takeover w/ the lock held, the other code locations use StopUnlocked(). A number of tests were calling Stop more than necessary. Improvements in the Client code make it unnecessary to call StopUnlocked() unless the particulars of the test require it, since Stop is automatic in all code paths.


mochi-co:
@jmacd Really appreciate all the work you've put into this! My apologises, I have been very busy lately - I'll try to look at this tonight, I'm excited to get the changes tested and released - it looks like you fixed a lot of things I overlooked :)

jmacd:
I'll be happy to add some new tests, there's no rush.

server/internal/clients/clients.go
```diff
@@ -217,29 +222,51 @@ func (cl *Client) ForgetSubscription(filter string) {
 
 // Start begins the client goroutines reading and writing packets.
 func (cl *Client) Start() {
-	cl.State.started.Add(2)
+	var started sync.WaitGroup
+
+	started.Add(2)
+	cl.State.endedR.Add(1)
```
jmacd, line 228: FYI I think these two Add(1) calls have to move above the corresponding `go` statements. This may contribute to #25, since without Add before the goroutine, it's possible for Stop to be called before one goroutine starts b/c Done can be reached w/o Add.

jmacd:
Please hold off on reviewing this. I would like to improve test coverage. I was also testing to see if #25 is completely fixed. When I tested before d9285d9, two of the tests would take 15s each to timeout, which was different than before this change.

d9285d9 fixed the test not to take 15s in those two places by not waiting for the writer goroutine to finish before closing the connection. However, if you don't wait for the writer, there's a new race potential around returning buffer to `s.bytepool` while it's still in use.

jmacd:
>  two of the tests would take 15s each to timeout, which was different than before this change.

This relates to holding the lock during Stop -- I'm not sure _why_ yet, but I'll resolve this.

PR #36, opened 2022-03-09
Move two WaitGroup.Add calls

jmacd:
Fixes most of #25. Extracted from #34.

The one (and only) test failure that I still see after this fix, also mentioned in #25, is:

```
--- FAIL: TestServerResendClientInflightBackoff (0.00s)
    server_test.go:2109: 
        	Error Trace:	server_test.go:2109
        	Error:      	Not equal: 
        	            	expected: 1
        	            	actual  : 2
        	Test:       	TestServerResendClientInflightBackoff
```

### DIFF
--- a/server/internal/clients/clients.go
+++ b/server/internal/clients/clients.go
@@ -218,6 +218,8 @@ func (cl *Client) ForgetSubscription(filter string) {
 // Start begins the client goroutines reading and writing packets.
 func (cl *Client) Start() {
 	cl.State.started.Add(2)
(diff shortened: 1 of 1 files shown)


jmacd:
I ran extensive tests on #38 and am confident that the fix here resolves test failures noted in this thread except those test failures that are due to (a) timing issues (see tests using Sleep()) and (b) reading an atomic variable w/o atomic function.

mochi-co:
This is really interesting. Great job spotting it, and such a simple fix.

Issue #37, opened 2022-03-13
Publish with retain doesn't retain


bldrdash:
Publish works fine except for honoring the retain flag.  Broker works fine with external publishing with retain.
Using master branch: 460f0ef6815ae482f85dc3b0b7ca01d789f70ad6

```golang
package main

import (
	"fmt"
	"os"
	"os/signal"
	"syscall"
	"time"

	mqtt "github.com/mochi-co/mqtt/server"
	"github.com/mochi-co/mqtt/server/listeners"
	"github.com/mochi-co/mqtt/server/listeners/auth"
)

func main() {
	sigs := make(chan os.Signal, 1)
	done := make(chan bool, 1)
	signal.Notify(sigs, syscall.SIGINT, syscall.SIGTERM)
	go func() {
		<-sigs
		done <- true
	}()

	server := mqtt.New()
	tcp := listeners.NewTCP("t1", ":1883")
	server.AddListener(tcp, &listeners.Config{Auth: new(auth.Allow)})

	// server.Serve doesn't block, so no need for goroutine, but for some
	// reason, we have to delay a LONG time for Serve's goroutines to start-up
	// > 1 second on a 5900X !!
	server.Serve()
	fmt.Println("\nServer is running")
	fmt.Println("Sleeping for 2 seconds")
	time.Sleep(2 * time.Second)

	// Publish with retain - doesn't work
	server.Publish("direct/publish", []byte("Published from broker"), true)
	fmt.Println("\nMessage published to \"direct/publish\" with retain")

	fmt.Println("\nTry running: mosquitto_sub -v -t \"direct/publish\"")

	<-done
}
```
```sh
go run main.go

# In another terminal window

# The following shows nothing
mosquitto_sub -v -t "direct/publish"

# But this works:
mosquitto_pub -t "mosquitto/publish" -m "from mosquitto_pub" -r
mosquitto_sub -v -t "mosquitto/publish"
```

Am I missing something?

wind-c:
trie.go:
// RetainMessage saves a message payload to the end of a topic branch. Returns
// 1 if a retained message was added, 0 if there was no change, and -1 if the
// retained message was removed.
func (x *Index) RetainMessage(msg packets.Packet) int64 {
	var q int64

	x.mu.Lock()
	defer x.mu.Unlock()
	n := x.poperate(msg.TopicName)
	if len(msg.Payload) > 0 {
		**if n.Message.FixedHeader.Retain == false {
			q = 1
		}**
[...]

PR #38, opened 2022-03-14
 Add an OnError handler, report the reason for disconnects

### LINKED ISSUE #21: client net.Conn in onConnect Hook

### LINKED ISSUE #33: Show the reason for client disconnects

jmacd:
Redo of #34.
Fixes #33.
Fixes #21.
Incorporates #36. 

Adds an OnError handler for errors that are not reported via the OnDisconnect handler.

Results from auditing error propagation in the server.
- an error was being skipped from ResendInflight
- an error was being ignored in `Reader.ReadFrom()`, which wasn't causing harm but made diagnosing other issues (and test coverage) more difficult

Adds a "Session reestablished" error to indicate when one client connection causes another to disconnect. For example, when two clients use the same ID, the server logs might read:

```
<< OnConnect client connected operator from [::1]:56779
<< OnDisconnect client disconnected operator connection from [::1]:56779: Session reestablished
```

Extends a few existing tests to hit test coverage.  Extensive testing reveals no new issues.

### DIFF
--- a/server/events/events.go
+++ b/server/events/events.go
@@ -1,12 +1,12 @@
 package events
 
 import (
-	"github.com/mochi-co/mqtt/server/internal/clients"
 	"github.com/mochi-co/mqtt/server/internal/packets"
 )
 
 type Events struct {
 	OnMessage    // published message receieved.
+	OnError      // server error.
 	OnConnect    // client connected.
 	OnDisconnect // client disconnected.
 }
@@ -15,15 +15,14 @@ type Packet packets.Packet
 
 type Client struct {
 	ID       string
+	Remote   string
 	Listener string
 }
 
-// FromClient returns an event client from a client.
-func FromClient(cl *clients.Client) Client {
-	return Client{
-		ID:       cl.ID,
-		Listener: cl.Listener,
-	}
+// Clientlike is an interface for Clients and client-like objects that
+// are able to describe their client/listener IDs and remote address.
+type Clientlike interface {
+	Info() Client
 }
 
 // OnMessage function is called when a publish message is received. Note,
@@ -44,3 +43,7 @@ type OnConnect func(Client, Packet)
 // is passed to the function if the client disconnected abnormally, otherwise it
 // will be nil on a normal disconnect.
 type OnDisconnect func(Client, error)
+
+// OnError is called when errors that will not be passed to
+// OnDisconnet are handled by the server.
+type OnError func(Client, error)
--- a/server/server.go
+++ b/server/server.go
@@ -4,6 +4,7 @@ package server
 import (
 	"errors"
 	"fmt"
+	"io"
 	"net"
 	"strconv"
 	"sync/atomic"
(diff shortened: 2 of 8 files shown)


server/server.go
```diff
@@ -166,6 +173,46 @@ func (s *Server) inlineClient() {
...
+	// If one of these ordinary conditions stops the connection,
+	// then the client closed or broke the connection.
+	if s.Events.OnError != nil &&
+		!errors.Is(err, io.EOF) {
+		s.Events.OnError(cl, err)
+	}
+	return err
+}
+
+func (s *Server) onStorage(cl events.Clientlike, err error) {
```
mochi-co, line 209: This is a nice addition

server/events/events.go
```diff
@@ -15,15 +15,14 @@ type Packet packets.Packet
...
 
-// FromClient returns an event client from a client.
-func FromClient(cl *clients.Client) Client {
-	return Client{
-		ID:       cl.ID,
-		Listener: cl.Listener,
-	}
+// Clientlike is an interface for Clients and client-like objects that
+// are able to describe their client/listener IDs and remote address.
+type Clientlike interface {
```
mochi-co, line 24: This is a good idea.

server/server.go
```diff
@@ -378,7 +421,7 @@ func (s *Server) Publish(topic string, payload []byte, retain bool) error {
 	}
 
 	if retain {
-		s.retainMessage(pk)
+		s.retainMessage(&s.inline, pk)
```
mochi-co, line 424: I wonder if this solves https://github.com/mochi-co/mqtt/issues/37 🤔

server/server_test.go
```diff
@@ -234,12 +276,8 @@ func TestServerEventOnConnect(t *testing.T) {
 	s, cl, _, _ := setupClient()
 	s.Clients.Add(cl)
 
-	var hookedClient events.Client
-	var hookedPacket events.Packet
-	s.Events.OnConnect = func(cl events.Client, pk events.Packet) {
-		hookedClient = cl
-		hookedPacket = pk
-	}
+	var hook packetHook
```
mochi-co, line 279: much tidier 👍🏻

mochi-co (approved):
Amazing work. I approve all these changes.

server/server.go
stffabi: Unfortunately this PR didn't fix #37, it seems like the retain flag is missing in the FixedHeader. I've created PR #42 which seems to fix the retain in my case.

PR #39, opened 2022-03-14
Wrap packet errors with cause information

jmacd:
Follow up from https://github.com/mochi-co/mqtt/pull/29#issuecomment-1053365833


mochi-co:
You put me to shame ;-) I will have endeavour to provide better error handling in the future. This is great!

PR #45, opened 2022-03-17
Adding Cluster Capabilities

wind-c:
A basic cluster is implemented using [hashicorp/memberlist](https://github.com/hashicorp/memberlist)

Cluster uses redis to store inflight and retain messages and subscriptions across nodes.

Please read the readme:examples/cluster/README.md


mochi-co:
@wind-c This is a nice surprise to wake up to! You've done a lot of work here. This is a fairly major addition so I am thinking it might be nice to get this into a 1.2.0 release.

I will take some time to go through this PR thoroughly - I invite recent contributors (@jphastings @rkennedy @jmacd @stffabi)  to review it also!

jmacd:
@wind-c I was curious about this PR and skimmed over it. I came to this repository looking for a broker to connect a single piece of hardware with a handful of sensors speaking SparkplugB/MQTT to an OpenTelemetry server (and I've nearly succeeded!). In that sense, I'm not looking for anything large-scale and in fact I could tolerate a brief downtime for what I'm doing, but more README text here would probably help me figure out how Clustering helps with my use-case. I think part of the problem is that "Cluster" doesn't describe a design in a specific way. Two things it appears to do are to give you a multi-homing capability (i.e., my device can connect to either broker) as well as a shared persistence layer. Because of the shared persistence, the two or more brokers would be effectively the same MQTT "instance". Is there a "leader" in this model? Assuming this picture is correct, it will be nice to summarize what kind of consistency losses or failure modes will result in from running in this mode, for example.

mochi-co (commented):
I like the inclusion of the examples file and benchmarks.

However, there are some elements which concern me:

1. The global replacement of the `packets` import with `packets2` and consequent usage, eg:
      ```go
      packets2 "github.com/mochi-co/mqtt/server/packets"
      ...
      func (cl *Client) WritePacket(pk packets2.Packet) (n int, err error) {...
      ```
    As far as I can tell this is a totally unnecessary, global change.

2. No comments on any types, funcs or vars. This will need to be resolved in order to maintain  the standards set forth in the [go report](https://goreportcard.com/report/github.com/mochi-co/mqtt). It would also be nice to see more comments explaining the intention behind the clustering mechanisms.

3. #40 introduces a redis persistence package. Having two separate but similar persistence packages makes me a little uncomfortable, but I could be convinced to go with it if the use case was clear and there are no other reasonable options.

4. Lots of changes to parts of the code outside of the `cluster` folder, but with no supporting context as to why these changes are being made (eg. the changes in `server/persistence` and the moving of the `internal/packets` package. 

I appreciate the work which has gone into this PR, and I think it provides some good ideas for where we could go for clustering support. However, in good faith I cannot review this PR further until the above issues are resolved.

PR #46, opened 2022-03-17
Subscribe: Only send retained messages if ACLs has allowed subscription to the topic

stffabi:


mochi-co (approved):
Very nice fix for a problem we didn't know existed :) Thank you @stffabi - this will go into v1.1.2!

stffabi:
Awesome, thanks so much for merging.

PR #51, opened 2022-03-17
Two no-functional-change cleanups combined

jmacd:
This combines:

1. Uses of `strconv.Itoa(int(x))` replaced with `strconv.FormatInt()` or `strconv.FormatUint()`
2. Tests that read values set by `atomic.StoreInt64()` use atomic reads as well.

The first would be needed to show system counters correctly on a 32-bit platform.
The second is part of #25 required to pass tests with `-race`.

### DIFF
--- a/server/internal/packets/packets.go
+++ b/server/internal/packets/packets.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"errors"
 	"fmt"
+	"strconv"
 )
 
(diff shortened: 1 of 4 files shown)


mochi-co (approved):
Sorry for the long wait in reviewing this! I like what you've done here, especially cleaning up the `&s.System.*` calls and also that everything is well tested!

This is a pleasant addition to v.1.2.0 :)

Issue #35, opened 2022-03-08
Do not disconnect existing session for new connection auth failures


jmacd:
In `Server.EstablishConnection()` there is an authorization test, and even if the test fails it first checks whether there's a session present and disconnects the other, valid session. This could lead to an invalid connection disrupting a valid connection.

It seems unintentional. Presumably MQTT does not require a correct response for `sessionPresent` when authorization has failed.

mochi-co:
Fixed in `v1.2.0` 👍🏻

PR #52, opened 2022-03-21
Server: Remove ClientState if the connection was a CleanSession session [MQTT-3.1.2-6]

stffabi:
This ensures the session state is not leaked and subscriptions
and buffer can be freed.

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -268,13 +268,7 @@ func (s *Server) EstablishConnection(lid string, c net.Conn, ac auth.Controller)
 		}
 		existing.Stop(fmt.Errorf("connection from %s: %w", cl.Info().Remote, ErrSessionReestablished))
 		if pk.CleanSession {
-			for k := range existing.Subscriptions {
-				delete(existing.Subscriptions, k)
-				q := s.Topics.Unsubscribe(k, existing.ID)
-				if q {
-					atomic.AddInt64(&s.System.Subscriptions, -1)
-				}
-			}
+			s.cleanSubscriptions(existing)
 		} else {
 			cl.Inflight = existing.Inflight // Take address of existing session.
 			cl.Subscriptions = existing.Subscriptions
@@ -330,6 +324,14 @@ func (s *Server) EstablishConnection(lid string, c net.Conn, ac auth.Controller)
 	}
 	err = cl.StopCause()
 
+	if pk.CleanSession {
+		// The state associated with a CleanSession MUST NOT be reused in any subsequent session [MQTT-3.1.2-6]
+		cl.Lock()
+		s.cleanSubscriptions(cl)
+		cl.Unlock()
+		s.Clients.Delete(cl.ID)
+	}
+
 	s.bytepool.Put(xbr) // Return byte buffers to pools when the client has finished.
 	s.bytepool.Put(xbw)
 
@@ -341,6 +343,17 @@ func (s *Server) EstablishConnection(lid string, c net.Conn, ac auth.Controller)
 	}
 
 	return err
+}
+
+// cleanSubscriptions cleans all the subscriptions of the client
+func (s *Server) cleanSubscriptions(cl *clients.Client) {
+	for k := range cl.Subscriptions {
+		delete(cl.Subscriptions, k)
+		q := s.Topics.Unsubscribe(k, cl.ID)
+		if q {
+			atomic.AddInt64(&s.System.Subscriptions, -1)
+		}
+	}
 }
 
 // writeClient writes packets to a client connection.


server/server.go
```diff
@@ -330,6 +324,14 @@ func (s *Server) EstablishConnection(lid string, c net.Conn, ac auth.Controller)
...
+	if pk.CleanSession {
+		// The state associated with a CleanSession MUST NOT be reused in any subsequent session [MQTT-3.1.2-6]
+		cl.Lock()
+		s.cleanSubscriptions(cl)
+		cl.Unlock()
+		s.Clients.Delete(cl.ID)
+	}
+
 	s.bytepool.Put(xbr) // Return byte buffers to pools when the client has finished.
 	s.bytepool.Put(xbw)
```
stffabi, line 336: For CleanSession=false sessions the `xbr` and `xbw` still can never be freed because the Client retains a pointer to those buffers. So if e.g. 100 clients were connected and all disconnect, there will be 200 buffers in the syncpool which never get freed.

jmacd (approved):

server/server.go
stffabi: It's not what I exactly meant by my comment. Deferring doesn't help to avoid the leaking of the buffers as long as the Client in the clients map have a pointer to them.

jmacd: I see. Somehow the Clients should be prevented from re-using those connections, I'd like to see them set `r` and `w` to `nil` after their reader and writer thread return, for example. Otherwise, I'd advocate not to use a sync.Pool, personally. I'm interested enough to take a look. 😁

stffabi: Exactly. What one does need to make sure is that there's also not another Goroutine (due to a publish) is still writing to the buffer in WriteBytes.

jmacd: Yeah, the package structure makes it not obvious when `WritePacket()` may be called from the server. I'd like to see more private control over writing to the connection. At least it should be clear (somehow) that if a Client disconnects that the server will not try to re-use its physical connection.

mochi-co: I've had a good look through the code and I don't see any real reason that defer shouldn't be used with the finalising `bytepool.Put` and `atomic.AddInt64` calls. If I recall correctly, I used sequenced calls here due to the performance cost of defer, however this cost was resolved in Go 1.4, so I see no standing reason to avoid this defer here (especially given that it's a relatively less frequent operation).

mochi-co: RE the possible leaking of xbr/xbw buffers:

I've been checking the logic of `EstablishConnection` to assess whether or not this is possible:

- xbr/xbw buffers are allocated with `bytepool.Get` whenever any new connection is initiated, regardless of it's validity.
- After a client connection has ended (either via EOF/error or disconnect), the xbr/xbw buffers are returned to the pool with `bytepool.Put`
- When a client connects with the same client-id, and and there client with the same id already registered, the new client is stopped and the existing clients subscriptions and inflight state are inherited. 
    - Note - #35 indicates an outstanding attack surface we should patch.
- The client connection state and buffers are not reused by the newly connected client.
- `client.Stop()` stops the buffer read/write loops, but it does not return the xbr/xbw buffers to the pool. 
- Additionally, any error throw before or during the Connect ACK process will fail to return the buffers to the pool, as the function returns before `bytepool.Put` can be run (this can be corrected with defer). A panic would also cause this to complete.

Any connected client which is disconnected in this manner will return the bytes to the pool (except as noted above) - the blocking`cl.Read(s.processPacket)` in `EstablishConnection` will unblock with an error, resulting in the client being stopped naturally and `bytepool.Put` being called. I believe this is the correct mechanism, as byte buffers should really only be mutable by their parent clients, and not by other goroutines.

As far as I understand it, implementing `defer` on the `bytepool.Put` calls should ensure no goroutine can end without first returning the buffers to the pool - any existing client goroutine will be ended as a result of `existing.Stop(..)` being called by the new connecting client, preventing the above scenario where buffers are leaked. 

Another possible solution would be to put the `bytepool.Put` calls into the `client.Stop` method, but I'm not convinced this is necessary.

Please sanity check this assumption and do correct me if you think I missed something!

stffabi: The problem is that the Client type still has a reference to the xbr/xbw and a client for which CleanSession=false will never be removed from `server.Clients`. Therefore the buffers are referenced from another path and will never be freed. As long as clients only connect, disconnect and reconnect that's not a big problem but consider the following scenario:

100 Clients successively connect to the server, because the `sync.Pool` is always empty all those client get 2 new buffers allocated by `bytepool.Get` calling the `New` function. All clients request sessions with `CleanSession=false`. Now all clients gracefully disconnect. So `EstablishConnection` will put those 2 buffers per client back into the pool with `bytepool.Put`. All 200 buffers with each 256KB in size will now be held by the sync.Pool. The server is now running idle and no more clients connect in near future. The sync.Pool is designed to release those buffers and if there's no other reference to the them they will be deallocated by the GC.

> // Any item stored in the Pool may be removed automatically at any time without
// notification. If the Pool holds the only reference when this happens, the
// item might be deallocated.

`Server.Clients` still holds a reference to every one of those 100 client structs, because of `CleanSession=false`, and every Client struct still have a pointer to the buffers they used. Therefore those 200 buffers won't be freed as long as the Clients are still registered in `server.Clients`.

mochi-co: Ah, I see what you are saying. I am envisioning something like this, perhaps:

```go
// ClearBuffers sets the read/write buffers to nil so they can be 
// deallocated automatically when no longer in use.
func (cl *Client) ClearBuffers() {
  cl.r = nil
  cl.w = nil
}
``` 

and then calling it after `cl.w.Stop()`:
https://github.com/mochi-co/mqtt/blob/0a7205e1104c323560d66bea8b8a83b38b31c19d/server/internal/clients/clients.go#L242-L258

And additionally deferred in `EstablishConnection` in order to handle deallocation when pre-read errors occur. 

Unless I am misunderstanding, this should remove any references to the buffers for disconnected clients and allow them to be deallocated. What do you think?

jmacd: Your `ClearBuffers` proposal makes sense and sounds like the right approach. I understand why Client objects are kept in memory after their connections are dead, AND I do not expect the server to use those connections again, only the associated state.

stffabi:
Seems like this PR still has some problems. It might now happen that for CleanSession=true a connection which has been forced disconnected (but not yet fully finished in EstablishConnection) can delete subscriptions of a new reconnected Client with the same ClientID.

So I'll close the PR, maybe someone could go ahead and fix those issues. Unfortunately I won't have time to investigate this further.

mochi-co:
@stffabi Thank you for your work on this, it has provided a great starting point for future improvements and sparked a useful discussion! 

I am happy to implement the aforementioned changes but it may be a little while as I will be busy for the next week or so.

server/server.go
jmacd: I looked at this a little more and it's not going to be a simple change. The clients code is organized around not having a lock, and I'm not sure how to avoid a race between Stop() on the one hand and the various calls to Write() on the other. If a mutex protected the `Done`, `r`, `w`, and `conn` variables this fix would be easier.

mochi-co: This is fixed in `v1.2.0` 👍🏻  It was a little tricky, but refactoring EstablishConnection helped me fix a lot of our outstanding issues. Thank you both for your guidance! 🙏🏻

PR #57, opened 2022-03-23
docker: add initial simple Dockerfile

deadprogram:
This PR adds a very basic Dockerfile base on Alpine and Go 1.18

To use it, first 

```
docker build -t mochi:latest .
```

Then to run it with TCP, Websocket, and also HTTP dashboard available on the host:

```
docker run -p 1883:1883 -p 1882:1882 -p 8080:8080 mochi:latest
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,31 @@
+FROM golang:1.18.0-alpine3.15 AS builder
+
+RUN apk update
+RUN apk add git
+
+WORKDIR /app
+
+COPY go.mod ./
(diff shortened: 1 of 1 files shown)


deadprogram:
Updated to use `ENTRYPOINT` instead of `CMD` so arguments can be passed in easily at runtime.

mochi-co:
Thanks @deadprogram - this is wonderful. I've tested it and it works well, so I'm merging it into the v1.2.0 branch :) Great to have #2 finally resolved.

PR #60, opened 2022-03-28
hotfix - system.info is not set correct back to listeners

anopid:
fix the issue with httpStat server, missing stats if you use BOLTDB
- added SetSystemStat(s *system.Info) to Listener Interface


mochi-co:
Hi @anopid - I'm so sorry, I managed to miss this PR originally and then kept making a note to reach out. If this is still an issue in v3.2.1, could you resubmit based on the current release with a description of the problem (or open an issue as it may be something with multiple solutions).

Thank you for your contribution and please accept my apologies for not getting back to you sooner. I appreciate your help!

PR #72, opened 2022-04-18
fix-panic

BoskyWSMFN:
fixed runtime panic in `server/internal/circ/pool.go` occurring on 32-bits architectures caused by misalignment of `BytesPool` struct members.

https://github.com/golang/go/issues/36606#issue-551005857

### DIFF
--- a/server/internal/circ/pool.go
+++ b/server/internal/circ/pool.go
@@ -7,8 +7,10 @@ import (
 
 // BytesPool is a pool of []byte.
 type BytesPool struct {
+	// int64/uint64 has to the first words in order
+	// to be 64-aligned on 32-bit architectures.
+	used int64 // access atomically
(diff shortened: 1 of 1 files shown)


mochi-co:
Good catch @BoskyWSMFN - I caused this in the v1.2.0 update. We need to find an easy way to check 32bit runtime...

Merging and releasing v1.2.1 👍🏻 Thank you!

BoskyWSMFN:
> Good catch @BoskyWSMFN - I caused this in the v1.2.0 update. We need to find an easy way to check 32bit runtime...
> 
> Merging and releasing v1.2.1 👍🏻 Thank you!

The easiest way is raspbian as far as I can see.
I compiled this broker for arm-32 system and caught the panic there so I can provide some feedback about compatability if I find something.

PR #92, opened 2022-08-17
Keep in sync server.System.Inflight

bkupidura:


bkupidura:
I dont like `require.Equal(t, s.System.Inflight, int64(-2))` in `server/server_test.go` - but all inflights was injected to server and not received. So s.System.Inflight is out-of-sync.

I see 3 options here:
* leave it as `-2` which looks strange :)
* remove this assertion from test
* set s.System.Inflight to proper value before clear

Any opinion?

mochi-co (approved):
This all looks great to me! 

I think -2 on the tests is just fine as it is very specifically testing subtraction, so it makes technical sense.

PR #96, opened 2022-09-02
Fix concurrent map iteration and map write

muXxer:
PR #90 introduced a regression.

If a client uses QoS 2, it happens that the broker panics with `fatal error: concurrent map iteration and map write`

```
fatal error: concurrent map iteration and map write

goroutine 16 [running]:
github.com/mochi-co/mqtt/server.(*Server).ResendClientInflight(0x1400015c000, 0x140000007e0, 0x0)
        mochi-mqtt/server/server.go:912 +0xf0
github.com/mochi-co/mqtt/server.(*Server).resendPendingInflights(0x1400015c000)
        mochi-mqtt/server/server.go:1122 +0x6c
github.com/mochi-co/mqtt/server.(*Server).eventLoop(0x1400015c000)
        mochi-mqtt/server/server.go:220 +0xf4
created by github.com/mochi-co/mqtt/server.(*Server).Serve
        mochi-mqtt/server/server.go:199 +0x70
```

This is because the `GetAll` method returns the map itself, instead of a copy.
The lock in `GetAll` doesn't help, because the map is used outside when the lock is already released.

There are two ways to fix that.
Either we hold the lock while iteration over the map (see `ForEach`), and use `SetWithoutLocking` and `GetWithoutLocking` in the loop, or we iterate over a copy of the map (see `ForEachCopy`).

I wasn't sure which approach is better, because one solution needs to copy the map, and the other may have more lock contention.

With the fix in this PR, at least the code doesn't panic anymore.

### DIFF
--- a/server/internal/clients/clients.go
+++ b/server/internal/clients/clients.go
@@ -49,46 +49,82 @@ func (cl *Clients) Add(val *Client) {
 	cl.Unlock()
 }
 
-// GetAll returns all the clients.
-func (cl *Clients) GetAll() map[string]*Client {
+// Copy returns a copy of the internal map of all the clients.
+func (cl *Clients) Copy() map[string]*Client {
 	cl.RLock()
 	defer cl.RUnlock()
-	return cl.internal
+
+	copiedClients := make(map[string]*Client)
+	for k, v := range cl.internal {
+		copiedClients[k] = v
+	}
+
+	return copiedClients
+}
+
+// ForEachCopy loops over a copy of the internal map of all the clients.
+// If the consumer function returns false, the iteration is stopped.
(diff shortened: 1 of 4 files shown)


mochi-co:
Hi @muXxer ! I was able to replicate it using inovex/mqtt-stresser.

Thanks very much for the PR! I found it quite interesting to read and see your approach, and it allowed me to see the issue more clearly. I think you have the general right idea about the problem - with your insights I was appear to have resolve the problem with very minimal changes (I think about 6 lines), so I will make an issue documenting the problem and PR those shortly.

mochi-co:
This issue has been resolved in https://github.com/mochi-co/mqtt/releases/tag/v1.3.2 
@muXxer Thanks again for your help in identifying the cause and solution this issue!

PR #97, opened 2022-09-02
Instead of waiting for the writing buffer to have enough space, skip writing and return an error

### LINKED ISSUE #95: Slow clients slow down the whole broker
We are using the MQTT broker and publishing messages directly to all clients using the broker's `Publish()` func.
This func adds a new publish packet to the `inlineMessages.pub` buffered channel (size 1024) and the `inlineClient()` loop will publish those packets to all subscribed clients.
For each subscribed client this will call `client.WritePacket()` which in the end will call `Write()` on the clients writer.

If a single subscribed client is too slow, the clients write buffer will fill up and the whole `inlineClient()` loop will hang until this client's buffer has space again (see `awaitEmpty` inside `Write()`). Shortly after the `inlineMessages.pub` buffered channel will fill up and further calls to `Publish()` will hang.

This means a single slow client (even one using QoS 0 with no guarantees of receiving packets) can make the whole broker wait indefinitely and not deliver any more packets to any client.

A possible workaround for this could be to instead of waiting for the buffer to be freed, to just return a "client buffer full" error and skip sending the packet to this client. If the client is using QoS 1/2 the inflight message retry mechanism should try to re-deliver the message.

What do you think? I can write a PR with this changes. Or do you have a better solution to this problem?

alexsporn:
This fixes #95

### DIFF
--- a/server/internal/circ/writer.go
+++ b/server/internal/circ/writer.go
@@ -1,9 +1,15 @@
 package circ
 
 import (
+	"errors"
 	"io"
 	"log"
 	"sync/atomic"
+)
+
+var (
+	// ErrWriterBufferFull indicates that there were not enough free bytes in the buffer to write.
+	ErrWriterBufferFull = errors.New("buffer is full, not enough space to write")
 )
 
 // Writer is a circular buffer for writing data to an io.Writer.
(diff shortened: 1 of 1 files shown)


mochi-co:
Let's discuss this in #95 and come to a complete solution before we approve 👍🏻

mochi-co (approved):

PR #105, opened 2022-09-29
set inflight when receive publish QoS>0 msg

cqcsdzmt:


server/server.go
```diff
@@ -423,6 +423,10 @@ func (s *Server) processPublish(cl *clients.Client, pk packets.Packet) error {
 			ack.FixedHeader.Type = packets.Pubrec
 		}
 
+		cl.Inflight.Set(pk.PacketID, clients.InflightMessage{
```
mochi-co, line 426: This will not work as the inflight message processor resends and replies to outgoing messages from the client, rather than incoming messages from the client. In effect, the publish packet will be retransmitted back to the sender when they reconnect.

PR #114, opened 2022-12-12
1. fix "func NewClient" and "func NewInlineClient" without options.

chenji1990:


mochi-co:
Hi @chenji1990! Thanks for the PR. I was just refactoring this early this morning as there are a couple of other areas this touches on that were overlooked (eg. maximum outbound quota), I will try to merge it shortly and release as v2.0.4

clients.go
```diff
@@ -171,8 +171,63 @@ func NewClient(c net.Conn, o *ops) *Client {
...
+	}
+
+	cl.refreshDeadline(cl.State.keepalive)
+
+	return cl
+}
+
 // NewInlineClient returns a client used when publishing from the embedding system.
-func NewInlineClient(id, remote string) *Client {
+func (s *Server) NewInlineClient(id, remote string) *Client {
```
mochi-co, line 205: I like the idea of add new client as a server method, however we shouldn't mix methods and files (eg. server method in client struct file) 👍🏻

chenji1990: I like your code and erudition. I will learn from you more in the future. If there is anything I can do for you, I will always be here.

mochi-co:
Closing as duplicated, by many thanks and fixed in v2.0.4 :)

Issue #115, opened 2022-12-13
Malformed packet errors on websocket listener


tommyminds:
Hi,

When publishing packages to the websocket listener using the Paho MQTT client library (both 3.11 and 5), I am getting many malformed packet errors. If I switch from ws to tcp listener, I am not getting these errors with the exact same client code.

<img width="1194" alt="Screenshot 2022-12-13 at 16 53 11" src="https://user-images.githubusercontent.com/123474/207380773-6bce7c38-1709-4fe3-abac-cadf7ed88611.png">

Any idea what the issue could be here?

mochi-co:
Thanks @tommyminds! I vaguely remember reading about a change in the accepted websocket protocol names, I look into it and see if i can replicate the issue!

tommyminds:
Unrelated, but it also seems the websocket library used by mochi has been put in archive mode. 

[https://github.com/gorilla/websocket](url)

Guess not really important right now, but might at some point also be worth looking to alternatives

mochi-co:
@tommyminds I haven't been able to replicate this with EMQX, would you be able to share a code sample? Thank you!

tommyminds:
I trimmed down the client code as much as I could to hopefully create something reproducable. I am running both code snippets against a local docker container running mochi-co. Not that I am not spreading out the message publishes (on purpose)

### Using the paho.golang (MQTT v5) client

```golang
package main

import (
	"context"
	"log"
	"net/url"
	"time"

	"github.com/eclipse/paho.golang/autopaho"
	"github.com/eclipse/paho.golang/paho"
)

func main() {
	ctx := context.Background()
	endpoint, _ := url.Parse("ws://localhost:1882/mqtt")

	client, err := autopaho.NewConnection(ctx, autopaho.ClientConfig{
		BrokerUrls: []*url.URL{endpoint},
		ClientConfig: paho.ClientConfig{
			ClientID: "test",
			OnClientError: func(err error) {
				log.Printf("server disconnected: %v", err)
			},
			OnServerDisconnect: func(d *paho.Disconnect) {
				log.Println("server disconnected")
			},
		},
		OnConnectError: func(err error) {
			log.Printf("connection error: %v", err)
		},
	})
	if err != nil {
		log.Fatalln(err)
	}

	defer client.Disconnect(ctx)

	publishTicker := time.NewTicker(1 * time.Second)
	defer publishTicker.Stop()

	for {
		<-publishTicker.C
[...]

PR #116, opened 2022-12-14
Fix websocket malformed packet bug

tommyminds:
Hi @mochi-co 👋🏻  Today I've spent some time with a colleague @svanharmelen, trying to understand the problem and search for the root cause.

We concluded that the problem is not with the client, as we tried to send the same amount of messages to Mosquitto (and also increased the test to 100.000 messages) which didn't give us any issues.

Next we looked at the websocket package that is used in your package and to rule things out we swapped it out for https://godoc.org/nhooyr.io/websocket. Funny enough this solved the issue, but as we were not convinced it's a bug in the Gorilla package and we didn't understand the root cause yet, we continued our search in `listeners/websocket.go`.

In there, we noticed in the [Read](https://github.com/mochi-co/mqtt/blob/master/listeners/websocket.go#L140) method that you ended with a [bufio.Read](https://pkg.go.dev/bufio@go1.19.4#Reader.Read) call and when looking up the documentation, we noticed that a single Read call does not guarantee to get all data up to the EOF.  

When we swapped to using Gorrila's [ReadMessage](https://github.com/gorilla/websocket/blob/v1.5.0/conn.go#L1091) method, we found out the problem was gone. In that method `ioutil.ReadAll` is being called on the net.Conn reader.

However, since [ioutil.ReadAll](https://cs.opensource.google/go/go/+/refs/tags/go1.19.4:src/io/io.go;drc=cb07765045aed5104a3df31507564ac99e6ddce8;l=694) creates a new buffer, and then keeps increasing it's size as need, that might impact performance. Thus we came up with a slightly more convoluted solution which looks like the following:

```golang
	var n, br int
	for {
		br, err = r.Read(p[n:])
		n += br
		if err != nil {
			if err == io.EOF {
				err = nil
			}
			return n, err
		}
	}
```

P.S. Following community best practices, we also removed the naked returns from the Read and Write methods on wsConn, but let us know if you want that to be reverted...

Fixes [#115](https://github.com/mochi-co/mqtt/issues/115)

### DIFF
--- a/listeners/websocket.go
+++ b/listeners/websocket.go
@@ -7,6 +7,7 @@ package listeners
 import (
 	"context"
 	"errors"
+	"io"
 	"net"
 	"net/http"
 	"sync"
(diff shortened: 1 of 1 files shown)


mochi-co:
Just a quick note to say I have seen this and it’s wonderful work, I will check it for consistency this evening and hopefully merge for a v2.0.5 release.

PR #123, opened 2022-12-18
Variable existed in the method processSubscribe is unstable

wind-c:
The variable existed can be changed repeatedly within a for loop. An array variable must be used to record the subscription of each filter.

### DIFF
--- a/server.go
+++ b/server.go
@@ -962,7 +962,7 @@ func (s *Server) processSubscribe(cl *Client, pk packets.Packet) error {
 		code = packets.ErrPacketIdentifierInUse
 	}
 
-	existed := false
+	existedMarks := make([]bool, len(pk.Filters))
 	reasonCodes := make([]byte, len(pk.Filters))
 	for i, sub := range pk.Filters {
 		if code != packets.CodeSuccess {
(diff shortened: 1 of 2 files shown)


server.go
```diff
@@ -978,16 +978,17 @@ func (s *Server) processSubscribe(cl *Client, pk packets.Packet) error {
...
 				atomic.AddInt64(&s.Info.Subscriptions, 1)
 			}
 			cl.State.Subscriptions.Add(sub.Filter, sub) // [MQTT-3.2.2-10]
 
 			if sub.Qos > s.Options.Capabilities.MaximumQos {
 				sub.Qos = s.Options.Capabilities.MaximumQos // [MQTT-3.2.2-9]
 			}
-
+			
+                        existedMarks[i] = !isNew
```
mochi-co, line 991: There's a tab formatting issue here, but I'll fix it post-merge 👍🏻

server.go
```diff
@@ -962,7 +962,7 @@ func (s *Server) processSubscribe(cl *Client, pk packets.Packet) error {
 		code = packets.ErrPacketIdentifierInUse
 	}
 
-	existed := false
+	existedMarks := make([]bool, len(pk.Filters))
```
mochi-co, line 965: I'm going to change this var name to `filterExisted` for clarity before release, but otherwise this is all a great improvement! 👍🏻 Thank you very much @wind-c!

mochi-co (approved):
This is a very effective changes which corrects an oversight in the original code.

PR #131, opened 2023-01-03
Save subscription properties for mqttv5

wind-c:
Otherwise the subscription properties will be lost during recovery.

### DIFF
--- a/hooks/storage/badger/badger.go
+++ b/hooks/storage/badger/badger.go
@@ -205,6 +205,12 @@ func (h *Hook) OnSubscribed(cl *mqtt.Client, pk packets.Packet, reasonCodes []by
 			Filter: pk.Filters[i].Filter,
 			Qos:    reasonCodes[i],
 		}
+		if pk.ProtocolVersion == 5 {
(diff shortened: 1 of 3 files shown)


hooks/storage/badger/badger.go
```diff
@@ -204,6 +204,12 @@ func (h *Hook) OnSubscribed(cl *mqtt.Client, pk packets.Packet, reasonCodes []by
 			Filter: pk.Filters[i].Filter,
 			Qos:    reasonCodes[i],
 		}
+		if pk.ProtocolVersion == 5 {
```
mochi-co, line 207: Sorry for the delay! I don't think we need a protocol version check here as the defaults for these values will be the same as the unset values in the struct, if they weren't set (i.e. they were protocol <5).

mochi-co (approved):
Great catch noticing this, I definitely missed it when putting it together. I'm going to merge it so you have the appropriate commit credit, and then move the values into the `in = &storage.Subscription{` statement as per my comment, before I issue the new release (to save time).

Thank you very much for your contribution! And apologies again for the delay.

PR #133, opened 2023-01-07
Change the order of inline client check

thedevop:
Change the order of inline client check to improve direct publish performance and avoid potential errors from custom ACLCheck.


mochi-co (approved):
Looks good to me, as discussed

PR #135, opened 2023-01-09
Improves Hooks thread safety

thedevop:
@mochi-co , current Hooks.internal is not thread safe. Although there is a Mutex lock when modifying, but reads do not acquire it. For this use case, changing internal to atomic.Value (which holds the slice) can provide the thread safety.

Further, if when Hooks is created ensure internal is stored with `h.internal.Store([]Hook{})`, then test for casting can be removed.
For example:
```
func (h *Hooks) OnConnect(cl *Client, pk packets.Packet) {
	i, ok := h.internal.Load().([]Hook)
	if !ok {
		return
	}
	for _, hook := range i {
		if hook.Provides(OnConnect) {
			hook.OnConnect(cl, pk)
		}
	}
}
```
can  be simplified to:
```
func (h *Hooks) OnConnect(cl *Client, pk packets.Packet) {
	for _, hook := range h.internal.Load().([]Hook) {
		if hook.Provides(OnConnect) {
			hook.OnConnect(cl, pk)
		}
	}
}
```

### DIFF
--- a/hooks.go
+++ b/hooks.go
@@ -110,7 +110,7 @@ type HookOptions struct {
 // Hooks is a slice of Hook interfaces to be called in sequence.
 type Hooks struct {
 	Log        *zerolog.Logger // a logger for the hook (from the server)
-	internal   []Hook          // a slice of hooks
+	internal   atomic.Value    // a slice of hooks
 	wg         sync.WaitGroup  // a waitgroup for syncing hook shutdown
 	qty        int64           // the number of hooks in use
 	sync.Mutex                 // a mutex
@@ -123,7 +123,11 @@ func (h *Hooks) Len() int64 {
 
 // Provides returns true if any one hook provides any of the requested hook methods.
 func (h *Hooks) Provides(b ...byte) bool {
-	for _, hook := range h.internal {
+	i, ok := h.internal.Load().([]Hook)
+	if !ok {
+		return false
+	}
+	for _, hook := range i {
 		for _, hb := range b {
 			if hook.Provides(hb) {
 				return true
@@ -138,16 +142,18 @@ func (h *Hooks) Provides(b ...byte) bool {
 func (h *Hooks) Add(hook Hook, config any) error {
 	h.Lock()
 	defer h.Unlock()
-	if h.internal == nil {
-		h.internal = []Hook{}
-	}
 
 	err := hook.Init(config)
 	if err != nil {
 		return fmt.Errorf("failed initialising %s hook: %w", hook.ID(), err)
 	}
 
-	h.internal = append(h.internal, hook)
+	i, ok := h.internal.Load().([]Hook)
+	if !ok {
+		i = []Hook{}
+	}
+	i = append(i, hook)
+	h.internal.Store(i)
 	atomic.AddInt64(&h.qty, 1)
 	h.wg.Add(1)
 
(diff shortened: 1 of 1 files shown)


mochi-co:
> current Hooks.internal is not thread safe. 

This is correct, I was not anticipating anyone adding and removing hooks after the server had begun processing connections 😅 I will try to look at this later this evening

mochi-co:
I had a quick look - As there are unlikely to be a huge number of hooks, and because the struct internals follow the same pattern as the other concurrent slices/maps, I think we can implement this more consistently and with more test-maintainability using the same GetAll pattern we use for Clients/Inflights/Subscriptions/Retained Messages/Topics/etc:

```go
// GetAll returns all the clients.
func (cl *Clients) GetAll() map[string]*Client {
	cl.RLock()
	defer cl.RUnlock()
	m := map[string]*Client{}
	for k, v := range cl.internal {
		m[k] = v
	}
	return m
}
```

For example:

```go
// GetAll returns all hooks.
func (h *Hooks) GetAll() []Hook {
	cl.RLock()
	defer cl.RUnlock()
	return append([]Hook{}, h.internal...)
}
```

and then replace:
```go
for _, hook := range h.internal {
```

with 

```go
for _, hook := range h.GetAll() {
```

Give this a go and see if it solves the issue. If it works, open a new PR and I'll see about getting it merged and released 👍🏻

mochi-co:
I have mixed feelings about my proposed solution above, the ore I think about it, the more I like the idea of using atomics instead of mutexes for this. I wonder if this can be extended to the other instances of GetAll to reduce copying.

I will look into this more later, hopefully. I think it's a very interesting proposition.

mochi-co:
I've run a series of tests to check the viability of a few different solutions. 

In regards to using sync.Atomic values to replace other instances where maps and slices are accessed concurrently:
- New data races occurred in the cases of maps - even if we can store and access the map atomically, the underlying memory addresses are still not thread-safe. For slices, we gained potential reduction in contention, but no memory benefit to storing the slices atomically as the data appears to be copied on load.
- I found no perceivable performance benefits under load using mqtt-stresser either in memory allocation or throughput. 

Regarding the changes to Hooks, the same findings applied. Because there's presently no case for adding and removing hooks while the server is running (and if there were, it would be a minimal write, high read scenario), we can gain the same benefit by using a RWMutex without the need for interface assertions, which are typically bad for compiler optimisations.

For these reasons, I think we should use the simple approach:

```go
func (h *Hooks) GetAll() []Hook {
	h.RLock()
	defer h.RUnlock()
	return append([]Hook{}, h.internal...)
}
+
for _, hook := range h.GetAll() {
```

I won't be available for the next few days so I am going to go ahead and fix this as I don't want to leave it broken. I will make a new PR and credit you as co-author in the commit as your work here has been instrumental. Finding this issue was a great help, and your solution was very well considered - I really appreciate your contributions, and it makes me very happy to see other people improving this code!

mochi-co:
Merged with commit credit in https://github.com/mochi-co/mqtt/pull/139/commits 👍🏻

thedevop:
@mochi-co , few things:

1. I agree that it is uncommon to add Hook after Serve.
3. I took advantage of atomicValue here because the content is a slice, and underlying content will not change, it will be append only. In case of reallocate, the old slice variable will still be valid (and consistent, but missing new entry), and new variable will be stored using atomic operation. Hence only Mutex is need for write, and no need for read.
4. This as you have pointed out, will not work with Map, unless you make a full copy of map during update, and store back the new map.
5. It is quite difficult to benchmark the difference. Based on my past experience, when < few hundred concurrency (not with long running lock), lock contentions are not too bad, but beyond that, it can be an issue. Given there are many hook point in the life cycle of a connection, when concurrent connections are high (we're looking to handle 200K+ per node), lock contentions become noticeable.
6. I appreciate keeping things consistent, it does make the code easier to understand/maintain. So there are 3 options:
   1. Use previous method, but make doc more clear from `AddHook attaches a new Hook to the server. Ideally, this should be called before the server is started with s.Serve().` --> `AddHook attaches a new Hook to the server. This MUST be called before the server is started with s.Serve(). `
   2. Use atomic.Value
   3. Use current method with RWMutex, you may consider changing `append([]Hook{}, h.internal...)` to `hooks := h.internal`. The latter will not copy the underlying array, it will just copy the slice structure.

PR #146, opened 2023-01-15
Fix inherited message sent before Connack

thedevop:


mochi-co:
This works, but I think there is a simpler implementation that will solve this issue - please see #147 and review

thedevop:
You're correct. I forgot the existing client is only deleted later on.

mochi-co:
Okay - I'll close this one and merge #147

Issue #143, opened 2023-01-14
Client can not connect if there are inherited retained messages


thedevop:
If there are retained messages and the client is not connecting with CleanSession, it will result retained messages sent before the Connack.

In the inheritClientSession, it will call publishRetainedToClient before it has a chance to sendConnack.
https://github.com/mochi-co/mqtt/blob/300152413c0e3a35741329068cfe12c5c502274f/server.go#L346

In the inheritClientSession, it can make a copy of the existing subscriptions, but process them after it completes sendConnack.

If you like, I can create a PR.

mochi-co:
Great discovery - I think you are correct. Yes, send over a PR and let's get it fixed! 

Thank you very much @thedevop

mochi-co:
Fixed in v2.1.6

PR #147, opened 2023-01-15
Publish retained messages only after connack

mochi-co:
Alternative to #143 / #144

- Ensure retained messages on inherited subscriptions are only published following the Connack packet.

### DIFF
--- a/server.go
+++ b/server.go
@@ -351,6 +351,13 @@ func (s *Server) attachClient(cl *Client, listener string) error {
 		return fmt.Errorf("ack connection packet: %w", err)
 	}
 
+	// Publish any retained messages for subscriptions which already existed on client takeover.
+	if sessionPresent {
+		for _, sub := range cl.State.Subscriptions.GetAll() {
+			s.publishRetainedToClient(cl, sub, true)
+		}
+	}
+
 	s.loop.willDelayed.Delete(cl.ID) // [MQTT-3.1.3-9]
 
 	if sessionPresent {
@@ -467,7 +474,6 @@ func (s *Server) inheritClientSession(pk packets.Packet, cl *Client) bool {
 				atomic.AddInt64(&s.Info.Subscriptions, 1)
 			}
 			cl.State.Subscriptions.Add(sub.Filter, sub)
-			s.publishRetainedToClient(cl, sub, existed)
 		}
 
 		return true // [MQTT-3.2.2-3]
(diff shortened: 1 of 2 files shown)


server.go
```diff
@@ -351,6 +351,13 @@ func (s *Server) attachClient(cl *Client, listener string) error {
 		return fmt.Errorf("ack connection packet: %w", err)
 	}
 
+	// Publish any retained messages for subscriptions which already existed on client takeover.
+	if sessionPresent {
+		for _, sub := range cl.State.Subscriptions.GetAll() {
```
mochi-co, line 356: As the new client cannot subscribe until the connack has been issued, GetAll can only contain inherited subscriptions.

server_test.go
```diff
@@ -445,7 +448,11 @@ func TestEstablishConnectionInheritExisting(t *testing.T) {
 		require.ErrorIs(t, v.StopCause(), packets.CodeDisconnect) // true error is disconnect
 	}
 
-	require.Equal(t, packets.TPacketData[packets.Connack].Get(packets.TConnackAcceptedSessionExists).RawBytes, <-recv)
+	connackPlusPacket := append(
+		packets.TPacketData[packets.Connack].Get(packets.TConnackAcceptedSessionExists).RawBytes,
+		packets.TPacketData[packets.Publish].Get(packets.TPublishRetain).RawBytes...,
+	)
+	require.Equal(t, connackPlusPacket, <-recv)
```
mochi-co, line 455: Ensure connack is always before retained message

mochi-co:
@thedevop

thedevop:
Looks good!

mochi-co:
@thedevop - I have a concerns about the use of `true` for the existed parameter. Please check my reasoning:

- A client subscription can be inherited from a new client from an existing client.
- A new client cannot subscribe before inheriting a subscription.
- Therefore, a new client cannot have any existing subscriptions.
- 1) Therefore, the value for existed in inheritClientSession must always be true, as the subscription already existed on the client being inherited from. 
- 2)  If so, then the logic `if !existed { atomic.AddInt64(&s.Info.Subscriptions, 1) }` in inheritClientSession must also be impossible to reach, and therefore serves no purpose.

Does this sound correct?

thedevop:
I didn't look into the Topics.Subscribe before, but you're correct, since these are inherited subscription from the same clientID, therefore it should always be existed.

Issue #145, opened 2023-01-15
Improves Hooks thread safety


thedevop:
Resurfacing this. @mochi-co , few things:

1. I agree that it is uncommon to add Hook after Serve.
3. I took advantage of atomicValue here because the content is a slice, and underlying content will not change, it will be append only. In case of reallocate, the old slice variable will still be valid (and consistent, but missing new entry), and new variable will be stored using atomic operation. Hence only Mutex is need for write, and no need for read.
4. This as you have pointed out, will not work with Map, unless you make a full copy of map during update, and store back the new map.
5. It is quite difficult to benchmark the difference. Based on my past experience, when < few hundred concurrency (not with long running lock), lock contentions are not too bad, but beyond that, it can be an issue. Given there are many hook points in the life cycle of a connection, when concurrent connections are high (we're looking to handle 200K+ per node), lock contentions become noticeable.
6. I appreciate keeping things consistent, it does make the code easier to understand/maintain. So there are 3 options:
   1. Revert to previous method (most performant) , but make doc more clear from `AddHook attaches a new Hook to the server. Ideally, this should be called before the server is started with s.Serve().` --> `AddHook attaches a new Hook to the server. This MUST be called before the server is started with s.Serve(). `
   2. Use atomic.Value
   3. Use current method with RWMutex, you may consider changing `append([]Hook{}, h.internal...)` to `hooks := h.internal`. The latter will not copy the underlying array, it will just copy the slice structure.

_Originally posted by @thedevop in https://github.com/mochi-co/mqtt/issues/135#issuecomment-1376545473_

mochi-co:
I have been thinking about it this week, and I agree with all your points. I've made pr #148 reflecting this changes, if you'd like to review it. I think there's also optimisations to be had in other key slices/maps.

mochi-co:
Fixed in v2.1.6

PR #148, opened 2023-01-15
Use Atomic instead of RWMutex for Hooks concurrency

mochi-co:
As per discussion in #145, converts RWMutex protection of Hooks slice to use atomic.Value instead.

cc @thedevop - please check and review 🙇🏻

### DIFF
--- a/hooks.go
+++ b/hooks.go
@@ -109,11 +109,11 @@ type HookOptions struct {
 
 // Hooks is a slice of Hook interfaces to be called in sequence.
 type Hooks struct {
-	Log          *zerolog.Logger // a logger for the hook (from the server)
-	internal     atomic.Value    // a slice of hooks
-	wg           sync.WaitGroup  // a waitgroup for syncing hook shutdown
-	qty          int64           // the number of hooks in use
-	sync.RWMutex                 // a mutex
+	Log        *zerolog.Logger // a logger for the hook (from the server)
+	internal   atomic.Value    // a slice of []Hook
+	wg         sync.WaitGroup  // a waitgroup for syncing hook shutdown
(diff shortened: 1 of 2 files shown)


hooks.go
```diff
@@ -110,7 +110,7 @@ type HookOptions struct {
 // Hooks is a slice of Hook interfaces to be called in sequence.
 type Hooks struct {
 	Log          *zerolog.Logger // a logger for the hook (from the server)
-	internal     []Hook          // a slice of hooks
+	internal     atomic.Value    // a slice of hooks
 	wg           sync.WaitGroup  // a waitgroup for syncing hook shutdown
 	qty          int64           // the number of hooks in use
 	sync.RWMutex                 // a mutex
```
thedevop, line 116: This can be changed to a Mutex lock.

hooks.go
```diff
@@ -136,18 +136,18 @@ func (h *Hooks) Provides(b ...byte) bool {
 
 // Add adds and initializes a new hook.
 func (h *Hooks) Add(hook Hook, config any) error {
-	h.Lock()
```
thedevop, line 139: Still needs a lock here for Add to prevent concurrent Add. Load/Store is atomic, but operations between them are not.

thedevop (changes requested):

hooks.go
mochi-co: Yes, I wondered about this as I was pushing it. This makes a lot more sense. Still, a little write lock is much better than RW locking.

mochi-co:
@thedevop changes have been made, if you are happy with it, I'll merge and it'll go into v2.1.6

thedevop (approved):
Thanks! Good to go!

PR #152, opened 2023-01-24
invalid config type provided

wind-c:
examples/persistence/bolt/main.go: invalid config type provided

### DIFF
--- a/examples/persistence/bolt/main.go
+++ b/examples/persistence/bolt/main.go
@@ -30,12 +30,15 @@ func main() {
 	server := mqtt.New(nil)
 	_ = server.AddHook(new(auth.AllowHook), nil)
 
-	err := server.AddHook(new(bolt.Hook), bolt.Options{
+	err := server.AddHook(new(bolt.Hook), &bolt.Options{
 		Path: "bolt.db",
 		Options: &bbolt.Options{
 			Timeout: 500 * time.Millisecond,
 		},
 	})
+	if err != nil {
+		log.Fatal(err)
+	}
 
 	tcp := listeners.NewTCP("t1", ":1883", nil)
 	err = server.AddListener(tcp)
--- a/server.go
+++ b/server.go
@@ -467,13 +467,11 @@ func (s *Server) inheritClientSession(pk packets.Packet, cl *Client) bool {
 			return false // [MQTT-3.2.2-3]
 		}
 
-		infLen := existing.State.Inflight.Len()
-		if infLen > 0 {
+		if existing.State.Inflight.Len() > 0 {
 			cl.State.Inflight = existing.State.Inflight // [MQTT-3.1.2-5]
 			if cl.State.Inflight.maximumReceiveQuota == 0 && cl.ops.capabilities.ReceiveMaximum != 0 {
 				cl.State.Inflight.ResetReceiveQuota(int32(cl.ops.capabilities.ReceiveMaximum)) // server receive max per client
 				cl.State.Inflight.ResetSendQuota(int32(cl.Properties.Props.ReceiveMaximum))    // client receive max
-				cl.State.Inflight.receiveQuota -= int32(infLen)                                // available receive quota is equal to the maximum quota minus the number of inflight messages
 			}
 		}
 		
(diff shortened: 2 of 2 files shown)


examples/persistence/bolt/main.go
```diff
@@ -30,12 +30,15 @@ func main() {
...
 	_ = server.AddHook(new(auth.AllowHook), nil)
 
-	err := server.AddHook(new(bolt.Hook), bolt.Options{
+	err := server.AddHook(new(bolt.Hook), &bolt.Options{
 		Path: "bolt.db",
 		Options: &bbolt.Options{
 			Timeout: 500 * time.Millisecond,
 		},
 	})
+	if err != nil {
```
mochi-co, line 39: Good find

server.go
```diff
@@ -467,7 +467,16 @@ func (s *Server) inheritClientSession(pk packets.Packet, cl *Client) bool {
 			return false // [MQTT-3.2.2-3]
 		}
 
-		cl.State.Inflight = existing.State.Inflight // [MQTT-3.1.2-5]
+		infLen := existing.State.Inflight.Len()
```
mochi-co, line 470: This is contrary to the mqtt v5 spec:
https://docs.oasis-open.org/mqtt/mqtt/v5.0/os/mqtt-v5.0-os.html#_Flow_Control

> 4.9 Flow Control [...] The send quota and Receive Maximum value are not preserved across Network Connections, and are re-initialized with each new Network Connection as described above. They are not part of the session state.

and 

> 3.1.2.11.3 Receive Maximum [...] The value of Receive Maximum applies only to the current Network Connection.

mochi-co (changes requested):
Thanks @wind-c ! 
We need the fix for the persistence example for sure - great job seeing it. However, we must not transfer send/receive quota state across connections as it goes against the v5 specification (see comments) and can cause a permanent client deadlock if no acks are returned.

If you can remove the inflight changes to server.go, I'll merge 👍🏻

wind-c:
You can test that when inflight messages are loaded from persistence without resetting quotas, an ErrReceiveMaximum  occurs when client publish message.

server.go
wind-c: `cl.State.Inflight = existing.State.Inflight`

This code overwrites the quota values of the client to zero, causing an ErrReceiveMaximum when the client publishes the message.

`  
func (s *Server) processPublish(cl *Client, pk packets.Packet) error {
	if !cl.Net.Inline && !IsValidFilter(pk.TopicName, true) {
		return nil
	}

	if atomic.LoadInt32(&cl.State.Inflight.receiveQuota) == 0 {
		return s.DisconnectClient(cl, packets.ErrReceiveMaximum) // ~[MQTT-3.3.4-7] ~[MQTT-3.3.4-8]
	}
`

server.go
```diff
@@ -467,7 +467,16 @@ func (s *Server) inheritClientSession(pk packets.Packet, cl *Client) bool {
...
+		if infLen > 0 {
+			cl.State.Inflight = existing.State.Inflight // [MQTT-3.1.2-5]
+			if cl.State.Inflight.maximumReceiveQuota == 0 && cl.ops.capabilities.ReceiveMaximum != 0 {
+				cl.State.Inflight.ResetReceiveQuota(int32(cl.ops.capabilities.ReceiveMaximum)) // server receive max per client
+				cl.State.Inflight.ResetSendQuota(int32(cl.Properties.Props.ReceiveMaximum))    // client receive max
+				cl.State.Inflight.receiveQuota -= int32(infLen)                                // available receive quota is equal to the maximum quota minus the number of inflight messages
```
mochi-co, line 476: Ah, sorry for the delay @wind-c - it's been a very busy week. I understand what you are trying to do now you have explained it, and I believe it is correct. The code looks good.

However, my one concern is still this line:

```go
cl.State.Inflight.receiveQuota -= int32(infLen)  // available receive quota is equal to the maximum quota minus the number of inflight messages`
```

Even though there are inflights still pending from the existing, the receive quota should be reset across connections (as specified in the spec). Therefore, we should remove this final line.

What do you think?

wind-c: ok, so remove this line.

mochi-co: @wind-c Yes, if this line is removed, I believe we can merge this PR. Apologies again for the delay in replying 🙏🏻

mochi-co (approved):
Looks good to me! Thank you for another excellent contribution, @wind-c

PR #155, opened 2023-02-01
Include a listener accepting an existing net.Listener

### LINKED ISSUE #154: Using an existing listener to serve incoming connections
Hi 👋, I have been playing around a bit with the MQTT protocol and would like to use an existing TCP net.Listener to serve incoming connections. This would allow for multiplexing over a single net listener. Is this a change you are willing to accept into your Repo? I could open MR to include the changes.

jeroenrinzema:
This MR includes the ability to serve incoming connection on an existing `net.Listener`.

resolves #154.

### DIFF
--- a/listeners/net.go
+++ b/listeners/net.go
@@ -1,0 +1,88 @@
+package listeners
+
+import (
+	"net"
+	"sync"
+	"sync/atomic"
+
+	"github.com/rs/zerolog"
+)
+
+// Net is a listener for establishing client connections on basic TCP protocol.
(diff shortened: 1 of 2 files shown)


jeroenrinzema:
The listener will only serve MQTT connections. The purpose of this change is that the listener could be defined outside of the MQTT package. Alternative listener implementations such as the [cmux](https://github.com/soheilhy/cmux) package could be used to multiplex a single net listener.

mochi-co (approved):
Looks good to me, thank you for your contribution @jeroenrinzema !

Issue #156, opened 2023-02-04
The client.refreshDeadline method should not be used within the client.WritePacket method. It will cause the device to be online for a long time even if it is offline


yemaozihecha:
```
// WritePacket encodes and writes a packet to the client.
func (cl *Client) WritePacket(pk packets.Packet) error {
	if atomic.LoadUint32(&cl.State.done) == 1 {
		return ErrConnectionClosed
	}

	if cl.Net.Conn == nil {
		return nil
	}

	defer cl.refreshDeadline(cl.State.keepalive)
	....
}
```

mochi-co:
Hi @yemaozihecha! This is an interesting find. Originally, the logic behind this is that the client deadline should be extended if the server successfully communicates/writes a packet to the client. However, in this case the defer statement means refreshDeadline is called even if the write failed. 

I will remove the defer statement and relocate the statement until after successful packet transmission.

Thank you!

mochi-co:
This should be resolved in v2.1.7

mochi-co:
The more I have thought about this, the more I am thinking we no longer need refreshDeadline in write, as we should be resetting it whenever we receive a ping request from the client. I may remove it in v2.2.0 which involves the changes for #150

yemaozihecha:
I think it should be used in client.ReadPacket, because the client's ping request can let you know whether the client is alive, but it is not needed in client.WritePacket.

mochi-co:
In v2.2.0 in #165 I have remove the refreshDeadline from client.WritePacket, as you are correct @yemaozihecha - hopefully it will be released soon, if not I will issue it as a separate minor release.

mochi-co:
This is resolved now in https://github.com/mochi-co/mqtt/releases/tag/v2.2.0 - thank you for highlighting it @yemaozihecha!

PR #157, opened 2023-02-04
Move refreshDeadline to only trigger on successful transmission

mochi-co:
As per #156, the refreshDeadline call has been moved so that it only extends the client deadline if packet transmission was successful. Previous the deadline would be extended even if the client was disconnected.


youmisun:
I think you should remove the refreshDeadline even if packet transmission was successful。Because I
use a device to test show that when i power off this device,  the packet also can transmission successful。

Issue #150, opened 2023-01-23
Purpose of FanPool


thedevop:
@mochi-co , can you please elaborate on the purpose of using FanPool? Currently when multiple publishers publish to same or different topics, FanPool will not ensure order nor provide concurrency protection (like PacketID issue).

mochi-co:
Hi @thedevop - sorry for the late reply on this, it's been a very busy week.

You may remember in v1.x.x, packet transmission was handled using a circular ring buffer (https://github.com/mochi-co/mqtt/tree/v1.1.1/server/internal/circ). It was fast and efficient, but it had some downsides which included dropping messages (#54, #109), and unusual race conditions and deadlocks (#101, #107). Additionally, each client required 3 goroutines to support this pattern (the client itself, a reader, and a writer), which was not suitable for everyone under high loads (#80).

However, more significantly, we found that (and this is slightly unrelated to the circular buffer) the dispatch pattern in v1 would cause the whole broke to stall if a connected client was slow in receipt (#95). This was one of the key motivators which forced me to actually get on with finishing v2, and then I had the perhaps naive idea to also upgrade us to MQTT v5 ("I'll just implement mqtt v5 while I'm making this change, how hard could it be").

I still needed some kind of go worker dispatch system, so I had a look at what others were doing. After a fair amount of benchmarking, I concluded that https://github.com/fhmq/hmq/blob/master/pool/fixpool.go was the most performant worker pool option; and in essence our fan pool is almost identical to this (as noted in the fanpool file header). 

During the development of v2, I switched off the fanpool entirely as I needed to monitor the broker internal processes without having to dig through goroutines. It wasn't until pre-release that I re-enabled it, and in doing so found no meaningful performance increase by having it enabled (at least in my environment). However, without the fanpool, we return to issue #95 - if a connected client is slow in receipt, the transmitting client would be delayed in sending messages to all subscribers. Moreover, the client (goroutines) became responsible for issuing messages to subscriptions, which meant publishing a message to a topic with thousands of subscribers would severely limit that client's own throughput, and so the fixed fan pool was re-enabled.

Having read your issue (#149), I am not convinced the fanpool is the correct solution. Neither am I 100% convinced that a goroutine for every worker is suitable either, although it is probably more flexible, and realistically anyone running a broker with hundreds of thousands of clients should probably have more than n*2kb in memory to spare. Especially if we can get clustering developed at some point (I know @bkupidura is working on some promising ideas).

[...]

Issue #149, opened 2023-01-23
Race condition for Client NextPacketID


thedevop:
Currently through FanPool, multiple messages from the the same publisher is processed by the same worker. However, this does not ensure messages from different publisher will be processed by the same worker. As a result, `(s *Server) publishToSubscribers` can be called concurrently, which in turn calls `s.publishToClient`, then calls `cl.NextPacketID()`, all of which can be triggered concurrently by different publishers. 

Although it atomically loads/stores packetID, the steps between these 2 operations are not atomic. So it is possible 2 publishers publishing to the same topic or different topics, can result in NextPacketID returning same PacketID to the same subscriber for different packets.
https://github.com/mochi-co/mqtt/blob/f90e52328d3abfebb9289441c18f86bcb245c2da/clients.go#L239

Furthermore, the following can corrupt the InFlight message, as it will just update the Inflight.
https://github.com/mochi-co/mqtt/blob/f90e52328d3abfebb9289441c18f86bcb245c2da/server.go#L831

Either a lock (probably more performant for the use case here) or CompareAndSwapUint32 can be used to eliminate the race condition.

Also, in a corner case, NextPacketID will get stuck in infinite loop if cl.State.packetID = 65535 and cl.State.Inflight contains 1-65535.

thedevop:
Race condition addressed in [PR 165](https://github.com/mochi-co/mqtt/pull/165)

PR #160, opened 2023-02-04
Replace fanpool with fixed worker pool and client write queues

mochi-co:
This PR addresses the issues highlighted in #150 and #149. 

In particular:
- Replaces the FanPool with a basic fixed-size worker pool.
- Client writes are now queued in packet channels per client. Instead of writing directly to the client, messages are added to the channels, which are then ranged by each client and written out in FIFO order.
- A lock has been added to NextPacketID.
- A mutex and lock has been added to Particle (topics), as I detected a race condition in my testing.

Interestingly, it improved the benchmark statistics, and in some cases considerably (see diff of readme). 

It is an operational proof-of-concept, in as much as it hasn't been tested on production, however, it passes PAHO reliably, and test coverage is maintained and passing. There may be improvements we can make to this code.

@thedevop - if could you look this over and give it a test, I would be very appreciative! Please let me know what you think.


server.go
```diff
@@ -26,10 +26,9 @@ import (
 )
 
 const (
-	Version                        = "2.1.7" // the current server version.
+	Version                        = "2.2.0" // the current server version.
 	defaultSysTopicInterval int64  = 1       // the interval between $SYS topic publishes
-	defaultFanPoolSize      uint64 = 32      // the number of concurrent workers in the pool
-	defaultFanPoolQueueSize uint64 = 1024    // the capacity of each worker queue
+	defaultPoolSize         uint64 = 8       // ...
```
mochi-co, line 31: I'll fix this comment later

server.go
```diff
@@ -1221,8 +1225,10 @@ func (s *Server) publishSysTopics() {
 func (s *Server) Close() error {
 	close(s.done)
 	s.Listeners.CloseAll(s.closeListenerClients)
-	s.fanpool.Close()
-	s.fanpool.Wait()
+	// s.fanpool.Close()
```
mochi-co, line 1228: need to remove these comments too

server_test.go
```diff
@@ -1944,6 +1964,7 @@ func TestServerProcessOutboundQos2Flow(t *testing.T) {
 	for i, tx := range tt {
 		t.Run("qos step"+strconv.Itoa(i), func(t *testing.T) {
 			r, w := net.Pipe()
+			time.Sleep(time.Millisecond)
```
mochi-co, line 1967: Had to add a few delays as there is a slight latency between adding a message to the queue and it being delivered

thedevop:
@mochi-co , few feedbacks:
1. Given each client has a buffered channel to receive writes (write chan). Is it better to have:
    1. `publishClient -> fixedPool (process/fan out to subscribers) -> subscriberWriteChan or`
    2. `publishClient (process/fan out to subscribers) -> subscriberWriteChan`
  Other than some lock contentions, everything should be non-blocking and process quickly.
2. One of the goal is not to prevent a slow receiving subscriber hold up publishing to other subscribers, so sending to the write chan needs to be non-blocking using select.
3. Potential Inflight issues (address separately from this PR?)
    1. Should there be a limit to max inflight msgs as well?
    2. Have you considered using a tree instead of a map so order can be retained?
    3. For QoS > 0, if the msg is dropped due to write chan full, then inflight needs to removed (or not set in the first place)

mochi-co:
> @mochi-co , few feedbacks:
> 1. Given each client has a buffered channel to receive writes (write chan). Is it better to have:
>    1. `publishClient -> fixedPool (process/fan out to subscribers) -> subscriberWriteChan or`
>    2. `publishClient (process/fan out to subscribers) -> subscriberWriteChan`
>       Other than some lock contentions, everything should be non-blocking and process quickly. In fact, most of the s.publishToClient can be done by the client write routine.

My concern with the second approach is that if Client A is publishing to subscribers of Topic, without an intermediary (pool) there's a possibility of blocking subsequent reads for Client A. Additionally, the hooks involved in this process are blocking, so in cases where a hook such as persistence is used, there could be a significant delay before these subsequent reads can occur. 

> 2. One of the goal is to prevent a slow receiving subscriber hold up publishing to other subscribers, so sending to the write chan needs to be non-blocking using select.

I'm not 100% certain what you mean in this case - could you elaborate?

> 3. Potential Inflight issues (address separately from this PR?)
>    1. Should there be a limit to max inflight msgs as well?

Inflights are currently limited by ReceiveMaximum on the client and server side. Is this what you mean?

>    2. Have you considered using a tree instead of a map so order can be retained?

I did consider this, but it becomes harder to select and delete/modify inflights by id. Currently the inflight map is simply sorted by created date before return. The alternative requires iterating through each inflight for any select/update/delete by id, and I wasn't able to determine a clean solution that performed significantly better than the existing map+output sort approach (also lock contention became an issue during iterations on larger datasets). As always, open to ideas, however.

>    3. For QoS > 0, if the msg is dropped due to write chan full, then inflight needs to removed (or not set in the first place)

Not 100% sure about this. I would expect that a dropped message would is the same as a network disruption - the QoS packet delivery should be retried when the client reconnects. As long as the QoS flow quota has not been exhausted, we should we not continue to queue inflights?

Thank you very much for your time! Look forward to hearing your (and anyone else's) thoughts! 🙂

thedevop:
> > 1. Given each client has a buffered channel to receive writes (write chan). Is it better to have:
> >    
> >    1. `publishClient -> fixedPool (process/fan out to subscribers) -> subscriberWriteChan or`
> >    2. `publishClient (process/fan out to subscribers) -> subscriberWriteChan`
> >       Other than some lock contentions, everything should be non-blocking and process quickly. In fact, most of the s.publishToClient can be done by the client write routine.
> 
> My concern with the second approach is that if Client A is publishing to subscribers of Topic, without an intermediary (pool) there's a possibility of blocking subsequent reads for Client A. Additionally, the hooks involved in this process are blocking, so in cases where a hook such as persistence is used, there could be a significant delay before these subsequent reads can occur.


From publishToSubscribers to publishToClient, the only hook is OnSelectSubscribers. If publishToClient is handled by client write routine, it would be behind a buffer channel and should offer the same level of protection as the fixed pool.


> > 2\. One of the goal is to prevent a slow receiving subscriber hold up publishing to other subscribers, so sending to the write chan needs to be non-blocking using select.
> 
> I'm not 100% certain what you mean in this case - could you elaborate?

Currently in publishToClient, send to cl.State.outbound is a blocking operation, it will not drop message to slow receiving subscriber per discussed. This can be changed to below so it's non-blocking:
```
	select {
	case cl.State.outbound <- out:
		atomic.AddInt32(&cl.State.outboundQty, 1)
	default:
		// call drop hook?
	}
``` 
    
> > 3\. Potential Inflight issues (address separately from this PR?)
> >     
> >     1. Should there be a limit to max inflight msgs as well?
> 
> Inflights are currently limited by ReceiveMaximum on the client and server side. Is this what you mean?

The following will directly add a message to inflight without any constraints, did I miss some other checks?
https://github.com/mochi-co/mqtt/blob/a8c9bdc48beffe3107b1bc4b8684639c3c7b1c5c/server.go#L836


> > 3\. For QoS > 0, if the msg is dropped due to write chan full, then inflight needs to removed (or not set in the first place)
> 
> Not 100% sure about this. I would expect that a dropped message would is the same as a network disruption - the QoS packet delivery should be retried when the client reconnects. As long as the QoS flow quota has not been exhausted, we should we not continue to queue inflights?

I was referring to intentionally dropping messages due to the subscriber is slow at receiving. Since subscriber was skipped due to write chan full, it should not exist in inflight neither.

mochi-co:
> From publishToSubscribers to publishToClient, the only hook is OnSelectSubscribers. If publishToClient is handled by client write routine, it would be behind a buffer channel and should offer the same level of protection as the fixed pool.

Ah, this makes me think we were talking about different things. Are you suggesting moving the enqueue from (for example):
```go
s.pool.Enqueue(func() {
    s.publishToSubscribers(pk)
})
```
to 
```go
for id, subs := range subscribers.Subscriptions {
   if cl, ok := s.Clients.Get(id); ok {
       s.pool.Enqueue(func() {
            _, err := s.publishToClient(cl, subs, pk)
	}   
    }
}
```
or are you suggesting to remove the fixed pool entirely?

---

> This can be changed to below so it's non-blocking:

This is very interesting, I've not seen this done for channel writes before. 

Is this in place of the following code (apologies, I don't have the ability to test is right now):
```go
if atomic.LoadInt32(&cl.State.outboundQty) >= s.Options.Capabilities.MaximumClientWritesPending {
    return out, packets.ErrPendingClientWritesExceeded
}
cl.State.outbound <- out
atomic.AddInt32(&cl.State.outboundQty, 1)
```
becomes
```go
select {
    case cl.State.outbound <- out:
        atomic.AddInt32(&cl.State.outboundQty, 1)
    default:
        cl.ops.hooks.OnPublishDropped(cl, pk) // perhaps new hook?
        return out, packets.ErrPendingClientWritesExceeded
}

```

cl.ops.hooks.OnQosDropped(cl, pk) should probably continue to only be used for inflight messages when they are dropped due to message expiry or clearing. This raises a question about renaming OnQosDropped to OnQosExpired for clarity.

---

> The following will directly add a message to inflight without any constraints, did I miss some other checks?

I see now what you are saying. The quota system restricts the number of messages that are 'inflight' but it does not restrict the number of messages that can be added to a client inflight queue, which is limited by the number of packet IDs available (65535). I think we could add a secondary restriction to the number of pending inflights per client, if you think it would be beneficial.

> Since subscriber was skipped due to write chan full, it should not exist in inflight neither.

Should this not be treated the same as QoS offline message queuing?

thedevop:
> Ah, this makes me think we were talking about different things. Are you suggesting moving the enqueue from (for example):
> 
> ```go
> s.pool.Enqueue(func() {
>     s.publishToSubscribers(pk)
> })
> ```
> 
> to
> 
> ```go
> for id, subs := range subscribers.Subscriptions {
>    if cl, ok := s.Clients.Get(id); ok {
>        s.pool.Enqueue(func() {
>             _, err := s.publishToClient(cl, subs, pk)
> 	}   
>     }
> }
> ```
> 
> or are you suggesting to remove the fixed pool entirely?

Yes, I was originally thinking of enqueue publishToClient. Now there are fixed pool, then there is also a buffered chan for the client write. I think we only need one or the other.



> This is very interesting, I've not seen this done for channel writes before.
> 
> Is this in place of the following code (apologies, I don't have the ability to test is right now):
> 
> ```go
> if atomic.LoadInt32(&cl.State.outboundQty) >= s.Options.Capabilities.MaximumClientWritesPending {
>     return out, packets.ErrPendingClientWritesExceeded
> }
> cl.State.outbound <- out
> atomic.AddInt32(&cl.State.outboundQty, 1)
> ```
> 
> becomes
> 
> ```go
> select {
>     case cl.State.outbound <- out:
>         atomic.AddInt32(&cl.State.outboundQty, 1)
>     default:
>         cl.ops.hooks.OnPublishDropped(cl, pk) // perhaps new hook?
>         return out, packets.ErrPendingClientWritesExceeded
> }
> ```

Yes. Here is an [example in Go Playground](https://go.dev/play/p/8JcJTvISjms).

> > The following will directly add a message to inflight without any constraints, did I miss some other checks?
> 
> I see now what you are saying. The quota system restricts the number of messages that are 'inflight' but it does not restrict the number of messages that can be added to a client inflight queue, which is limited by the number of packet IDs available (65535). I think we could add a secondary restriction to the number of pending inflights per client, if you think it would be beneficial.

You're right, it does have 65K bound, so we don't need another check. But more important, for message that was dropped in the 1st place, it should not make into the inflight queue.

> 
> > Since subscriber was skipped due to write chan full, it should not exist in inflight neither.
> 
> Should this not be treated the same as QoS offline message queuing?

I would treat them differently. It can be confusing that msg was not sent in the 1st place, but then all the sudden show up upon reconnect. Messages will also be out of order. For example, msg X is dropped when write chan is full, then msg Y subsequently came in and write chan is free. X will be delivered later on upon reconnect. In addition, given the plan is to call OnPublishDropped when msg is drooped, upper layer may have already taken corrective action, and can conflict when msg is delivered again later on.

mochi-co:
Closing in lieu of new PR

thedevop:
Race condition address in [PR 165](https://github.com/mochi-co/mqtt/pull/165)

PR #162, opened 2023-02-05
failed to delete inflight data

wind-c:
The s.hooks.OnQosPublish method needs to be called, otherwise the following s.hooks.OnQosComplete or processPuback(s.hooks.OnQosComplete) method will report a data not found error.


mochi-co (approved):
This makes sense

mochi-co:
Released in v2.1.8 - thank you very much @wind-c !

PR #163, opened 2023-02-08
Add Clone to system.Info

thedevop:


mochi-co (approved):
Great idea, this will go out shortly in the next release!

PR #165, opened 2023-02-09
Client write buffers

mochi-co:
Following our discussion in #150 and #160 

This PR removes the fixed pool in exchange for writing directly to the clients via client-specific buffered write channels. In the event that a client's write buffer is full, in such an event as a slow reading client, the message is dropped and the new OnPublishDropped hook is called. If the message was QoS 1 or 2, the inflight is rolled back and the message is treated as unattempted.

Also corrects tests and updates benchmarks. Paho is passing.

cc @thedevop thank you for your insights! If you could take a look and see if this PR is appropriate, I would be very grateful :) My apologies for the delay!


clients.go
```diff
@@ -237,6 +250,8 @@ func (cl *Client) refreshDeadline(keepalive uint16) {
 // If no unused packet ids are available, an error is returned and the client
 // should be disconnected.
 func (cl *Client) NextPacketID() (i uint32, err error) {
+	cl.Lock()
+	defer cl.Unlock()
 	i = atomic.LoadUint32(&cl.State.packetID)
 	started := i + 1
```
thedevop, line 256: Need a check if started is > 65535, otherwise the following for loop and be infinite if Inflight is full.

thedevop (changes requested):
Looks good, just one inherited bug that's not related to this change.

clients.go
mochi-co: Are you sure?
For started to be >= 65535, then it relies on cl.State.packetID being 65534/65535 +1.

If i is 65535, it will immediately step into the overflow logic and reset i to 1. If i is 65534, it will increment and use 65535 as the next packet id if it is not already in use. The following packet id request will follow the aforementioned logic and return to 1. If we then cycle all the way back to our start index without finding a vacant id, then it will fail with a packet ids exhausted (ErrQuotaExceeded) error (see the TestClientNextPacketID* tests for some examples).

As yet I was not able to put it into an endless loop - however, if you are able to reproduce a loop, definitely please open a new issue for it! 

For now, I will get this change merge and released as v2.2.0 and we can address the above after if necessary 😄 

Thank you for your assistance! I've been busy lately and it's wonderful to get a second pair of eyes on this stuff - I really appreciate it a lot.

PR #169, opened 2023-02-10
Fix potential NextPacketID endless loop, expand tests

mochi-co:
Fixes potential endless loop scenario with clients.NextPacketID and updates related tests to cover this scenario.
cc @thedevop Thank you for your code sample - it really helped illustrate the issue!

### DIFF
--- a/clients.go
+++ b/clients.go
@@ -252,7 +252,12 @@ func (cl *Client) refreshDeadline(keepalive uint16) {
 func (cl *Client) NextPacketID() (i uint32, err error) {
 	cl.Lock()
 	defer cl.Unlock()
+
 	i = atomic.LoadUint32(&cl.State.packetID)
+	if i >= 65535 {
+		i = 0
+	}
+
 	started := i + 1
 	overflowed := false
 	for {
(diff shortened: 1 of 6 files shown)


thedevop (changes requested):
This works. However, there is a slightly optimized solution:
Change:
```
	i = atomic.LoadUint32(&cl.State.packetID)
	if i >= 65535 {
		i = 0
	}

	started := i + 1
```
to
```
	i = atomic.LoadUint32(&cl.State.packetID)
	started := i
```
With `started := i + 1`, started will be checked twice.

[Example with](https://go.dev/play/p/g8kFzu7RjkY) `started := i`
[Example with](https://go.dev/play/p/R0ObHan6Dwd) `started := i + 1`

mochi-co:
@thedevop I believe this optimization will put the method into an infinite loop if all packet ids are allocated, as started can be 65535 but `if overflowed && i == started {` can never be reached.

mochi-co:
Optimization was implemented by refactoring NextPacketID, and small changes were made to enhance ease of testing (introduction of unexported Capabilities.maximumPacketID field). Additionally, max uint16 values were replaced with math constants.

PR #176, opened 2023-02-22
Use *packets.Packet for outbound chan

### LINKED ISSUE #175: Large memory consumption with high concurrent connections
On a node that is running with ~140K concurrent connections, using 2.1.6 it was consuming ~8GB of memory. However, after upgrade to 2.2.1, the application crash with OOM with ~100K using ~20+GB.

The culprit is:
https://github.com/mochi-co/mqtt/blob/1e8f92210202b61beb7f9fd104b057ec0a8386db/clients.go#L160

I was originally expecting only 8Bytes * MaximumClientWritesPending per client. However, after examining this more closely, this is using `chan packets.Packet` instead of `chan *packets.Packet`, which will use significantly more memory. Since this is last step in publishToClient, and no more modification will be made to the Packet either from this function or callers of publilshToClient, we can make outbound `chan *packets.Packet`.

thedevop:
This is the minimum change needed to use *packets.Packet for outbound chan. I do suggest more use of point to packets.Packet for internal calls, as it will result in less copy/memory allocation, for example cl.WritePacket. Fix #175.

### DIFF
--- a/server_test.go
+++ b/server_test.go
@@ -1408,7 +1408,7 @@ func TestPublishToClientExceedClientWritesPending(t *testing.T) {
 	s.Clients.Add(cl)
 
 	for i := int32(0); i < cl.ops.capabilities.MaximumClientWritesPending; i++ {
-		cl.State.outbound <- packets.Packet{}
+		cl.State.outbound <- &packets.Packet{}
 		atomic.AddInt32(&cl.State.outboundQty, 1)
 	}
 
--- a/clients.go
+++ b/clients.go
@@ -135,17 +135,17 @@ type Will struct {
 
 // State tracks the state of the client.
 type ClientState struct {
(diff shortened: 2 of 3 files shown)


server_test.go
```diff
@@ -1408,7 +1408,7 @@ func TestPublishToClientExceedClientWritesPending(t *testing.T) {
 	s.Clients.Add(cl)
 
 	for i := int32(0); i < cl.ops.capabilities.MaximumClientWritesPending; i++ {
-		cl.State.outbound <- packets.Packet{}
+		cl.State.outbound <- &packets.Packet{}
```
mochi-co, line 1411: I think new(packets.Packet) would work better here for consistency

mochi-co (approved):
This is great, thank you for opening it! Everything passes as expected so I'll merge it in. Incidentally I noticed a very marginal performance increase in the benchmarks.

Issue #173, opened 2023-02-21
reconnection problem when use same client id 


werbenhu:
BUG: If there is already a connection with a client ID, and another connection uses the same client ID to connect again, the new client cannot receive the subscription message, and the hook should call OnDisconnect() first when the new connection comes. , and then call OnConnect()

thedevop:
Ensure the original client is not set to autoReconnect, otherwise both clients will take turn to connect and kick off the other client.

I just tried it with client not set to autoReconnect, and the new client with the same clientID is able to receive the subscription message.

@mochi-co , it is more logical to have OnDisconnect before OnConnect. But tricky given the current structure. BTW, in looking into this, I found cl.WritePacket is now no longer called by a single thread, so there can be multiple writes to the connection simultaneously.

werbenhu:
我已收到你的邮件，我将在看到邮件后，尽快给你回复。

werbenhu:
@thedevop Neither client has set to autoReconnect, but both of their Clean Session are set，you can try it again with Clean Session on.

thedevop:
Thanks @werbenhu. I was able to reproduce the issue.

Below is a description of the issue:
1. Each connection is handled by its own Go routine, and stays in attachClient
2. When a new connection comes in with the same clientID, and let's assume for the rest of this discussion the previous connection (session) has cleanSession flag set, it will disconnect the previous connection (send out a disconn packet and potentially forcefully disconnect the connection) and unsubscribe all previous session topics, but it does not wait until the previous Go routine to exit
3. It will then add the new session
4. At a later time, the previous connection Go routine will detect the connection is lost, then it does a clean up
https://github.com/mochi-co/mqtt/blob/1e8f92210202b61beb7f9fd104b057ec0a8386db/server.go#L367-L371
5. This will remove the subscriptions of the previous session (normally not an issue as the new session probably didn't receive subscription yet), and delete the this clientID from Server.Clients (this is an issue, from the broker's perspective, this client doesn't exist any more)

In fact, if a 3rd session is created with the same clientID, it will now exhibit a different issue, both 2nd and 3rd session connections will exist on the broker, and broker state only thinks there is 3rd session.

@mochi-co , there are few potential solutions to this, for example, add an additional state, or move the clean up into cl.State.endOnce in cl.Stop, etc...

werbenhu:
[...]

PR #179, opened 2023-02-23
reconnection problem when use same client id #173

werbenhu:


mochi-co:
Thank you for the PR @werbenhu! In the end the issue ended up being a big more complicated than expected, and I had to make some significant changes, so I will close this PR - but thank you for your work!

PR #184, opened 2023-03-01
Add Prometheus metrics

thedevop:
@mochi-co , I had to do `go vendor` to add modules in go.mod. So you may want to ignore go.mod and go.sum and regenerate them on your side.


system/system.go
```diff
@@ -51,11 +56,83 @@ func (i *Info) Clone() *Info {
 		MessagesDropped:     atomic.LoadInt64(&i.MessagesDropped),
 		Retained:            atomic.LoadInt64(&i.Retained),
 		Inflight:            atomic.LoadInt64(&i.Inflight),
-		InflightDropped:     atomic.LoadInt64(&i.InflightDropped),
-		Subscriptions:       atomic.LoadInt64(&i.Subscriptions),
-		PacketsReceived:     atomic.LoadInt64(&i.PacketsReceived),
-		PacketsSent:         atomic.LoadInt64(&i.PacketsSent),
-		MemoryAlloc:         atomic.LoadInt64(&i.MemoryAlloc),
-		Threads:             atomic.LoadInt64(&i.Threads),
+		//		InflightDropped:     atomic.LoadInt64(&i.InflightDropped),
```
mochi-co, line 59: Is this supposed to be commented?

system/system.go
```diff
@@ -51,11 +56,83 @@ func (i *Info) Clone() *Info {
...
+		{"g", "clients_connected", "A gauge of number of currently connected clients", &i.ClientsConnected},
+		{"g", "clients_disconnected", "A gauge of total number of persistent clients", &i.ClientsDisconnected},
+		{"c", "clients_maximum", "A count of maximum number of active clients that have been connected", &i.ClientsMaximum},
+		{"g", "clients_total", "A gauge of total number of connected and disconnected clients with a persistent session currently connected and registered", &i.ClientsTotal},
+		{"c", "messages_received", "A counter of total number of publish messages received", &i.MessagesReceived},
+		{"c", "messages_sent", "A counter of total number of publish messages sent", &i.MessagesSent},
+		{"c", "messages_dropped", "A counter of total number of publish messages dropped to slow subscriber", &i.MessagesDropped},
+		{"g", "retained", "A gauge of total number of retained messages active on the broker", &i.Retained},
+		{"g", "inflight", "A gauge of the number of messages currently in-flight", &i.Inflight},
+		//		{"c", "inflight_dropped", "A", &i.InflightDropped},
```
mochi-co, line 92: As above

system/system.go
```diff
@@ -51,11 +56,83 @@ func (i *Info) Clone() *Info {
...
+					},
+					fn,
+				),
+			)
+		}
+	}
+
+	buildInfo := prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			// Namespace: AppName,
```
mochi-co, line 130: As above

listeners/http_sysinfo.go
```diff
@@ -68,6 +69,9 @@ func (l *HTTPStats) Init(log *zerolog.Logger) error {
 
 	mux := http.NewServeMux()
 	mux.HandleFunc("/", l.jsonHandler)
+	mux.Handle("/metrics", promhttp.Handler())
+
+	// mux.Handle("/metrics", promhttp.)
```
mochi-co, line 74: Also this 😄

mochi-co:
Overall this looks good! However, I wonder if it would be possible to implement it as a hook, as not everyone will want to overhead of prometheus in the main code.

If we used`OnSysInfoTick`, we just need a way to it to the http sys info listener. Which makes me wonder if that should be hook based too...

system/system.go
thedevop: I forgot to mention, it appears InflightDropped is not used any more, I can't find any reference where the broker updates it.

system/system.go
thedevop: same as above, InflightDropped is not updated

system/system.go
thedevop: I can't find any reference as to what's the app name you call this. mqtt, mochicomqtt, etc? can you provide one?

thedevop:
I was originally thinking creating a sub-module, but since all its data are coming from the System, so it may make sense just be part of that.

The way I registered the metrics, there is really no activities on it unless there is a pull on the handle. And upon pull, it does atomic load of the values.

I was thinking of putting as a config into the sysinfo, but that config only contains TLS and is shared with many other listeners. As a result I hardcoded into the sysinfo listener. I'm open to suggestion how you would like to have this implemented:
1. implement at server level, add config to server Options
2. implement at sysinfo listener, use sysinfo specific config
3. as a hook, but i'm not sure if it's appropriate, as it really is just a one time registration, and connect to a HTTP handler

listeners/http_sysinfo.go
thedevop: that was a mistake... 😄

thedevop:
I'll resubmit a PR in the future.

PR #192, opened 2023-03-14
Add Prometheus metrics

oliverpool:
Since #184 has been abandoned, here is my take on adding Prometheus metrics.

It exposes something like:
```
mochi_co_mqtt_info{version="2.2.6",goversion="go1.20.2"} 1
started 1678799214
time 1678799670
uptime 456
bytes_received 0
bytes_sent 0
clients_connected 0
clients_disconnected 0
clients_maximum 0
clients_total 0
messages_received 0
messages_sent 0
messages_dropped 0
retained 0
inflight 0
inflight_dropped 0
subscriptions 0
packets_received 0
packets_sent 0
memory_alloc 2998272
threads 7
```

I made a couple of decisions, which could be challenged:
- don't add external dependencies (since the metrics format is quite simple, I don't think that this is needed)
- use reflection, so that any change to the `system.Info` struct is automatically exposed on `/metrics`
- export a `WriteMetrics` method (since I intend to push the metrics to VictoriaMetrics, instead of pulling)

### DIFF
--- a/listeners/http_sysinfo.go
+++ b/listeners/http_sysinfo.go
@@ -7,9 +7,11 @@ package listeners
 import (
 	"context"
 	"encoding/json"
+	"fmt"
 	"io"
 	"net/http"
 	"reflect"
+	"runtime"
 	"strconv"
 	"sync"
 	"sync/atomic"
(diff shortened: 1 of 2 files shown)


oliverpool:
@thedevop maybe you have any feedback, since you authored #184

thedevop:
Thanks @oliverpool. This is nice way of exporting metrics in Prometheus format without the use of Prometheus library, and should be suffice for people that uses the standalone application. However, for those that embed this module in their own application, most likely they're already using Prometheus library if they're exporting Prometheus metrics, and #184 integrates well with that.

@mochi-co , it comes down to if you prefer not using Prometheus library for standalone broker app or a unified solution for those that embed this as a module.

bkupidura:
I agree with @thedevop here, there are cases (majority?) when mqtt server is just a part of bigger piece, which probably already have metrics exposed.
I would just add mqtt metrics as part of prometheus lib, and user can extend this with his custom metrics.

mochi-co:
This PR is somewhat out of date now. I would recommend that we discuss it as a group and co-author @oliverpool if any of these changes are applied. I will close the PR for now - however note that it's not an indication of the code quality or functionality, which is otherwise very good. @oliverpool I'm sure we would all be grateful if you provided your experience in future discussions around this topic in https://github.com/orgs/mochi-mqtt/discussions/280#discussioncomment-6755593

PR #195, opened 2023-03-21
Simplified code

wind-c:
Simplify the code for the loadInflight and loadRetained methods. Adjust the validation order of the processSubscribe method to ensure that it fails quickly if there is an error, since s.hooks. OnACLCheck generally takes a long time.


mochi-co:
This looks good! I feel like msgToPacket should really be a method on storage.Message, but I'm going to merge it and make the change after. Apologies for the delay!

Issue #78, opened 2022-06-11
The OnConnect and the OnDisconnect event hook trigger order is wrong when using the same clientid


youmisun:
my implement code
```
        server.Events.OnConnect = func(cl events.Client, pk events.Packet) {
		fmt.Printf("<< OnConnect client connected clientid=%s\n", cl.ID)
	}

	server.Events.OnDisconnect = func(cl events.Client, err error) {
		fmt.Printf("<< OnDisconnect client disconnected clientid=%s, err=%v\n", cl.ID, username, err)
	}
```
client 1 connects to the server, it prints
```
<< OnConnect client connected clientid=mqttx_f988b2da
```
client 2 use the same clientid as client 1 connets to the server, it prints
```
<< OnConnect client connected clientid=mqttx_f988b2da
<< OnDisconnect client disconnected clientid=mqttx_f988b2da err=client session re-established
```
i think when client 2 use the same clientid as client 1 connets to the server, it should print like
```
<< OnDisconnect client disconnected clientid=mqttx_f988b2da err=client session re-established
<< OnConnect client connected clientid=mqttx_f988b2da
```
it should be trigger the OnDisconnect before the OnConnect

mochi-co:
I've reviewed the code associated with the OnConnect and OnDisconnect events and I don't think it's reasonably possible to swap the sequence of these events, as they are triggered as they occur in different goroutines with no cross-communication.

Your best option may be to track the client ids on the event handling side and then only execute the relevant code if the client has been reconnected - eg. `server.ErrSessionReestablished`

PR #198, opened 2023-03-28
#78 storage hook should not execute the relevant code if the client has been reconnected

werbenhu:
storage hook should not execute the relevant code if the client has been reconnected.


werbenhu:
我已收到你的邮件，我将在看到邮件后，尽快给你回复。

mochi-co (approved):
Nice catch! This looks good to me. Thank you for adding tests too :)

Issue #206, opened 2023-05-02
Payload Format Indicator in last will cause malformed packet error


sukvojte:
**Description:**
If client sends `Payload Format Indicator` in last will message (which is part of `CONNECT` message), then connection fails on `read 
connection: malformed packet: will properties` error.

According to `OASIS MQTT Version 5.0 Standard` is `Payload Format Indicator` allowed in `Will Properties` field of `CONNECT` message (more info in chapter `3.1.3.2.3 Payload Format Indicator` of standard)

**Affected version**
mochi-co/mqtt - v2.2.7

**Client**
ESP32-WROOM32u with ESP-IDF v5.0.1 - MQTT5 example

**Message example (wireshark dump)**
```
MQ Telemetry Transport Protocol, Connect Command
    Header Flags: 0x10, Message Type: Connect Command
    Msg Len: 121
    Protocol Name Length: 4
    Protocol Name: MQTT
    Version: MQTT v5.0 (5)
    Connect Flags: 0xee, User Name Flag, Password Flag, Will Retain, QoS Level: At least once delivery (Acknowledged deliver), Will Flag, Clean Session Flag
    Keep Alive: 120
    Properties
    Client ID Length: 12
    Client ID: ESP32_xxxxxx
    Will Properties
        Total Length: 38
        ID: Will Delay Interval (0x18)
        Value: 10
        ID: Payload Format Indicator (0x01)
        Value: 1
        ID: Publication Expiry Interval (0x02)
        Value: 10
        ID: Response Topic (0x08)
        Length: 14
        Value: /test/response
        ID: Correlation Data (0x09)
        Length: 6
        Value: 123456
    Will Topic Length: 11
    Will Topic: /topic/will
    Will Message Length: 12
    Will Message: 692077696c6c206c65617665
    User Name Length: 3
    User Name: 123
    Password Length: 3
    Password: 456
```

**Proposed solution**
Allow `Payload Format Indicator` in `CONNECT` message:

```
diff --git a/packets/properties.go b/packets/properties.go
index 6c8d56b..da809b2 100644
--- a/packets/properties.go
+++ b/packets/properties.go
@@ -42,7 +42,7 @@ const (
 
 // validPacketProperties indicates which properties are valid for which packet types.
 var validPacketProperties = map[byte]map[byte]byte{
-	PropPayloadFormat:          {Publish: 1},
+	PropPayloadFormat:          {Connect: 1, Publish: 1},
[...]

PR #207, opened 2023-05-02
Fix packet validation according to OASIS MQTT Version 5.0 - 3.1.3.2.3…

sukvojte:
Proposed fix of #206


mochi-co:
You're absolutely right on this - great find. However, the problem extends a little further than this single property, so I've made a PR #210 which correctly validates the encoding and decoding of will properties. I will note you as a co-author for your contribution!

PR #216, opened 2023-05-06
Expire session if SessionExpiryInterval is 0

thedevop:
If SessionExpiryInterval was not set in CONNECT, SessionExpiryIntervalFlag is also not set. According to spec:
```If the Session Expiry Interval is absent the value 0 is used. If it is set to 0, or is absent, the Session ends when the Network Connection is closed.```


mochi-co:
Great catch. I will merge this immediately 👍🏻 Thanks as always, @thedevop !

Issue #221, opened 2023-05-06
Remove close of outbound chan


thedevop:
@mochi-co, now that there is a client context, we can use that to exit the WriteLoop, and no longer need to close the outbound channel. I also took advantage of that so we can remove use of lock in cl.Stop.

I haven't created a PR yet as I haven't updated all the unit tests (the ones in hooks/storage is a bit tricky as they don't construct clients using newClient). Can you review [this commit](https://github.com/thedevop/mqtt/commit/fb33974ab1885990a47ebfc1332fec3400c87821) and let me know if that's the direction we should take. If so, can you also suggest how you like to update the testClient in unit tests in hooks/storage.

mochi-co:
I've left a few comments - generally I think this is the right direction. 

If you keep the nil checks in Stop() and Closed() it should work without any major changes to the test logic. As I noted - we should keep these checks as client.State is exported, and if a developer creates a client without setting these it will panic when the client disconnects.

On this note perhaps it would make sense to export the open/ctx/cancel fields too. If we were to do this, then I would probably suggest renaming State.cancel to State.CancelCtx so it's more clear what it relates to.

Otherwise looks solid!

thedevop:
Sure I'll added in the nil check. However, although client.State is exported, it is meant for the developer some read access, and not for developer to actually create/write to top level items. As currently, if a developer create/set a new ClientState, this will "break" as there are many critical parts that are not exported, like outbound chan...

I can make both ctx/cancel exported to make unit test from other modules easier, but feels a bit "dangerous", as  I don't see a use case where the context is canceled and the rest of the client is not properly shutdown.

Cancel was stored in Client as where it is created is not where it will be called. We can not subsequently change the ctx as WriteLoop is in a different Go routine.

Let me know if you prefer ctx/cancel named as open/cancelOpen or ctx/cancelCtx?

thedevop:
I added in the nil check, most of the test cases do not need to be modified anymore. I renamed ctx/cancel to open/cancelOpen, feel free to change these names.

mochi-co:
I've merged and released this as v2.2.10 🥳 Thank you very much @thedevop

Issue #202, opened 2023-04-22
Unsafe close of outbound channel


yh742:
Hi there,

Found a possible issue while running test (related to #193). It seems like closing the outbound channel can cause race condition since other publisher could be publishing to it at the same time.

```
WARNING: DATA RACE
Write at 0x00c00048c070 by goroutine 21:
  runtime.closechan()
      /usr/local/go/src/runtime/chan.go:357 +0x0
  github.com/mochi-co/mqtt/v2.(*Client).Stop.func1()
      /home/sean/go/pkg/mod/github.com/mochi-co/mqtt/v2@v2.2.7/clients.go:368 +0x10f
  sync.(*Once).doSlow()
      /usr/local/go/src/sync/once.go:74 +0x101
  sync.(*Once).Do()
      /usr/local/go/src/sync/once.go:65 +0x46
  github.com/mochi-co/mqtt/v2.(*Client).Stop()
      /home/sean/go/pkg/mod/github.com/mochi-co/mqtt/v2@v2.2.7/clients.go:358 +0x74
  github.com/mochi-co/mqtt/v2.(*Server).processDisconnect()
      /home/sean/go/pkg/mod/github.com/mochi-co/mqtt/v2@v2.2.7/server.go:1150 +0x1bb
  github.com/mochi-co/mqtt/v2.(*Server).processPacket()
      /home/sean/go/pkg/mod/github.com/mochi-co/mqtt/v2@v2.2.7/server.go:561 +0x787
  github.com/mochi-co/mqtt/v2.(*Server).receivePacket()
      /home/sean/go/pkg/mod/github.com/mochi-co/mqtt/v2@v2.2.7/server.go:411 +0x87
  github.com/mochi-co/mqtt/v2.(*Server).receivePacket-fm()
      <autogenerated>:1 +0x89
  github.com/mochi-co/mqtt/v2.(*Client).Read()
      /home/sean/go/pkg/mod/github.com/mochi-co/mqtt/v2@v2.2.7/clients.go:349 +0x19c
  github.com/mochi-co/mqtt/v2.(*Server).attachClient()
      /home/sean/go/pkg/mod/github.com/mochi-co/mqtt/v2@v2.2.7/server.go:366 +0x9c4
  github.com/mochi-co/mqtt/v2.(*Server).EstablishConnection()
      /home/sean/go/pkg/mod/github.com/mochi-co/mqtt/v2@v2.2.7/server.go:312 +0x88
[...]

Issue #191, opened 2023-03-13
OnConnectAuthenticate fail cause  cl.WriteLoop memory leaks.


likangjia:
the client verification failed because cl.State.outbound is not closed, resulting in the cl.WriteLoop block cannot be closed. How to handle memory leaks？

github.com/mochi-co/mqtt/v2/server.go:303
`

        if !s.hooks.OnConnectAuthenticate(cl, pk) { // [MQTT-3.1.4-2]

		err := s.sendConnack(cl, packets.ErrBadUsernameOrPassword, false)

		if err != nil {

			return fmt.Errorf("invalid connection send ack: %w", err)

		}

		return packets.ErrBadUsernameOrPassword
	}

	.......

	s.Log.Debug().Str("client", cl.ID).Err(err).Str("remote", cl.Net.Remote).Str("listener", listener).Msg("client disconnected")
	expire := (cl.Properties.ProtocolVersion == 5 && cl.Properties.Props.SessionExpiryIntervalFlag && cl.Properties.Props.SessionExpiryInterval == 0) || (cl.Properties.ProtocolVersion < 5 && cl.Properties.Clean)
	s.hooks.OnDisconnect(cl, err, expire)
	close(cl.State.outbound)`

thedevop:
We can consider putting `close(cl.State.outbound)` in `cl.Stop`.

likangjia:
At present, due to device migration or other reasons, the authentication of the device may always fail and cannot be completed correctly

PR #222, opened 2023-05-08
Use context to exit WriteLoop

thedevop:
This is the PR for #221. By using context to exit WriteLoop, we no longer need to close the outbound chan.

This also impacts:
* #202 
* #191


mochi-co (approved):

mochi-co:
This looks great to me, thank you very much @thedevop !

PR #231, opened 2023-05-18
Fix example usage of NewHTTPStats

plourdedominic:
Update the example file to avoid confusion for future users


mochi-co (approved):
Good catch!

Issue #233, opened 2023-05-23
Reopen #153 (Possible issue with Will retain flag)


dadebue:
### This is just a placeholder issue

From my experience the problem with LWT messages still exists. I've added details in #153.
Can you please reopen the issue #153? This issue can be deleted then...

BR
Marco

dgduncan:
I have noticed the same thing so I am going to take a look at this when I get a moment.

dgduncan:
@dadebue I opened up a PR that should resolve this issue.

mochi-co:
Merged @dgduncan's PR and released as v2.2.12 - let me know if it resolves the issue :) Thanks for finding this!

dadebue:
I can happily confirm that the issue is resolved! Thanks to everyone :)

dgduncan:
@dadebue Glad to hear it! Let me know if you notice anything strange. I have a strong use case for LWT and retained LWT that has not been implemented yet so if problems still persist I would love to know :)

Issue #153, opened 2023-01-28
Possible issue with Will retain flag


dgduncan:
When testing, I discovered that there may be an issue with the retain flag when Will is published. When setting the Will to retain true it doesn't seem to actually set the retain flag to true when it is published. I can see the message when subscribed to the topic when it is published; however, the message will not be there when subscribed again.

mochi-co:
Thanks @dgduncan - I will try to look into this as soon as I can. What MQTT protocol/client are you using?

dgduncan:
@mochi-co paho.mqtt.golang 3.1.1
I will attempt to do some investigation myself as it was late when I posted this.

dgduncan:
I am beginning to wonder if this isn't an LWT problem but perhaps a retain flag problem. I have run a few tests where I publish multiple message from the same client to a topic:


	token2 = mqttclient.Publish("LWT/admin", 1, true, "connected")
	fmt.Println(token2.WaitTimeout(60 * time.Second))

	token2 = mqttclient.Publish("LWT/admin", 1, true, "connected")
	fmt.Println(token2.WaitTimeout(60 * time.Second))

And when I subscribe to that topic:

```
[DEBUG] [client]   Connect()
[DEBUG] [store]    memorystore initialized
[DEBUG] [client]   about to write new connect msg
[DEBUG] [client]   socket connected to broker
[DEBUG] [client]   Using MQTT 3.1.1 protocol
[DEBUG] [net]      connect started
[DEBUG] [net]      received connack
[DEBUG] [client]   startCommsWorkers called
[DEBUG] [client]   client is connected/reconnected
[DEBUG] [net]      incoming started
[DEBUG] [net]      startIncomingComms started
[DEBUG] [net]      outgoing started
[DEBUG] [net]      startComms started
[DEBUG] [pinger]   keepalive starting
[DEBUG] [client]   startCommsWorkers done
[DEBUG] [store]    enter Resume
[DEBUG] [store]    exit resume
[DEBUG] [client]   exit startClient
[DEBUG] [client]   enter Subscribe
[DEBUG] [client]   SUBSCRIBE: dup: false qos: 1 retain: false rLength: 0 MessageID: 1 topics: [LWT/admin]
[DEBUG] [client]   sending subscribe message, topic: LWT/admin
[DEBUG] [net]      outgoing waiting for an outbound message
[DEBUG] [net]      logic waiting for msg on ibound
[...]

PR #234, opened 2023-05-30
Add retainMessage to LWT to properly handle message retention

dgduncan:
As mentioned in both #153 and #233 the retain header was useless as no downstream code utilized it. This issue was duplicated using MQTTX for both LWT and LWT with a DelayInterval.  s.retainMessage is only ever called while processing a Publish packet and thusly is not calling during the processing of the LWT.

### DIFF
--- a/server.go
+++ b/server.go
@@ -1285,6 +1285,10 @@ func (s *Server) sendLWT(cl *Client) {
 		return
 	}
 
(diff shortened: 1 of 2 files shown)


dgduncan:
Added missed retain flag on broken test.

mochi-co (approved):
This looks good to me! Thank you for fixing it :)

PR #237, opened 2023-06-12
Add OnRetainPublished hook

thedevop:
Add OnRetainPublished hook, it is called when a retained message is published.


mochi-co:
Interesting use case! Looks good to me so I'll add it. I do love a good pull request.

PR #239, opened 2023-06-14
Update topics.go

x20080406:
fix issue. topic "a/b/c" can be subscribe by "a/b"


mochi-co:
Hi @x20080406 - could you please provide a reproducible test case as I have not seen this occur so far... thanks!

x20080406:
My fix is invalid. it's still has a bug. the root cause is :client1 subscribes to topic ‘a/b’, client2 sends a message to topic ‘a/b/c’, theoretically client1 should not receive messages from client2, but actually received messages from client2.

PR #247, opened 2023-06-25
Add OnSessionEstablish hook

dgduncan:
This PR adds the OnSessionEstablish hook. The reasoning behind this addition is to begin to add functionality that will enable future "cluster" mode work. I.E this allows the server an opportunity to check for client data on a remote store and pull in that client before going through before inheriting the session or creating a new client session. 

This would allow for a different type of persistence topology where clients are imported on a per client connection basis and not a "big bang" import as is current done in the current persistence implementations.  The current topology works well for a single server type design and not a multi server or multi node type design.


mochi-co:
This looks good to me, I'll add it to the next release! Sorry for the delay in reviewing!

mochi-co:
> this allows the server an opportunity to check for client data on a remote store and pull in that client before going through before inheriting the session or creating a new client session.

It may be interesting to migrate to this model entirely instead of loading all clients at the beginning. I can't think of many situations where this wouldn't be a superior pattern.

PR #248, opened 2023-06-26
[Delay Merging] Migrate from zerolog to slog

dgduncan:
This PR, if approved, will be merged upon the release of GO v1.21.

This PR changes the repository over from `zerolog` to `slog`. Some important considerations were taken when working on this PR:

- `zerolog` is capable of low to zero allocation logging. Because of this, I decided to use `LogAttrs` to get as close to this performance as possible at the expense of readability.
- `slog` does not have any equivalent to `.Err()` where you can pass an `error` directly to and it will check if `nil` in `.Err()` itself. To get around this, an `if` has been included on those calls and the log lines all but duplicated. I would love some feedback and see if anyone has any other ideas for this.


dgduncan:
I am a bit confused about coveralls does their line calculation. How could there be a 0.5% change when lines were only replaced?

dgduncan:
I have removed the Delay Merging tag in the name as I believe now with the release of GO `1.21` we can move forward.

torntrousers:
Big change to review but LGTM. Will this go in with a Mochi minor version update as some indication people need to also update Go in ther environment?

dgduncan:
@torntrousers I believe. I was also thinking if there was any apatite for some more experimental version tag to denote that this is a big change and a way to test before having it in an official version release; however, I don't want to start muddying up our versions.

werbenhu:
I've reviewed the changes, and everything looks fine. LGTM. I looked into slog, and it doesn't directly support interface-based abstraction. Perhaps we can make it more flexible by creating an adapter to achieve a similar interface-like behavior. This interface could include common log functions such as Info(), Infof(), Error(), Errorf(), etc. We could provide a default logger implementation using slog, and users could switch to other loggers by implementing their own logger structs and swapping them in the Server code. @mochi-co @dgduncan, I'd like to hear your thoughts on this approach.

dgduncan:
@werbenhu If I understand what you mean correctly, I personally feel that providing a `Logger` type interface that can be interchanged is of lower importance than moving more towards using the standard library. If there is an expressed interest by users for that functionality we can work to implement.

werbenhu:
@dgduncan I've written a simple stress test to compare the performance differences between slog.LogAttrs(), slog.Info(), and zerolog.Info(). Below is my code and the results:
```
package main

import (
	"context"
	"os"
	"testing"

	"github.com/rs/zerolog"
	"golang.org/x/exp/slog"
)

func BenchmarkSlogInfo(b *testing.B) {
	f, err := os.OpenFile("a.txt", os.O_CREATE|os.O_WRONLY|os.O_TRUNC, 0600)
	if err != nil {
		panic(err)
	}
	defer f.Close()

	log := slog.New(slog.NewTextHandler(f, nil))
	b.ResetTimer()
	for n := 0; n < b.N; n++ {
		log.Info(
			"hello world",
			"index", n,
			"name", "werben",
			"age", 30,
		)
	}
}

func BenchmarkSlogAttr(b *testing.B) {
	f, err := os.OpenFile("b.txt", os.O_CREATE|os.O_WRONLY|os.O_TRUNC, 0600)
	if err != nil {
		panic(err)
	}
	defer f.Close()

	log := slog.New(slog.NewTextHandler(f, nil))
	b.ResetTimer()
	ctx := context.TODO()
	for n := 0; n < b.N; n++ {
		log.LogAttrs(ctx,
			slog.LevelInfo,
			"hello world",
			slog.Int("index", n),
			slog.String("name", "werben"),
			slog.Int("age", 30),
		)
	}
}

func BenchmarkZerologInfo(b *testing.B) {
	f, err := os.OpenFile("c.txt", os.O_CREATE|os.O_WRONLY|os.O_TRUNC, 0600)
	if err != nil {
		panic(err)
	}
	defer f.Close()

	zlog := zerolog.New(os.Stderr).Output(zerolog.ConsoleWriter{Out: f})
	b.ResetTimer()

	for n := 0; n < b.N; n++ {
		zlog.Info().
			Int("index", n).
			Str("name", "werben").
			Int("age", 30).
			Msg("hello world")
	}
}
```
cpu: Intel(R) Core(TM) i5-7400 CPU @ 3.00GHz
BenchmarkSlogInfo-4                95996             11695 ns/op               7 B/op          0 allocs/op
BenchmarkSlogAttr-4               103012             10719 ns/op               0 B/op          0 allocs/op
BenchmarkZerologInfo-4             61846             17071 ns/op            2169 B/op         65 allocs/op

From my results, slog outperforms zerolog by a significant margin. However, the difference between slog.LogAttrs() and slog.Info() is not very significant. Personally, I lean towards the style of log.Info(), which is concise and elegant. I'm unsure about how others perceive this code style. If the majority share my sentiments, could we consider sacrificing a slight amount of performance for code that is more elegant and easier to comprehend?

mochi-co:
I'm just moving house this week so I'm a bit short on connectivity, but I'm going to reply to all these shortly - it's great work you've done!

dgduncan:
@werbenhu Wow thank you so much for doing that test and putting some numbers to what I have read. I absolutely am in favor of a more elegant and easier to read `Info` compared to `LogAttrs` if the performance is indeed this close.

dgduncan:
@werbenhu @mochi-co I have replaces all instances of `LogAttrs` with their equivalent `Info`, `Error`, etc. 

Opinion question, in the `main.go` files I currently left the `log.Fatal(err)` as the equivalent would be something along the `server.Log.Error("", "error", err.Error()` followed by `os.Exit(1)`. How do you two feel about replacing these logs with the new slog format or keeping them there?

thedevop:
@dgduncan, there maybe a 3rd option, slog can be wrapped into a struct which implements Fatal.
```go
type Logger struct {
	*slog.Logger
}

func (s *Logger) Fatal(msg string, args ...any) {
	s.Error(msg, args...)
	os.Exit(1)
}
```

@werbenhu , I expanded your benchmark to include outputting in json, which is where zerolog shines:
```go
func BenchmarkSlogInfoJson(b *testing.B) {
	f, err := os.OpenFile("aa.txt", os.O_CREATE|os.O_WRONLY|os.O_TRUNC, 0600)
	if err != nil {
		panic(err)
	}
	defer f.Close()

	log := slog.New(slog.NewJSONHandler(f, nil))
	b.ResetTimer()
	for n := 0; n < b.N; n++ {
		log.Info("hello world", "index", n, "name", "werben", "age", 300)
	}
}

func BenchmarkSlogAttrJson(b *testing.B) {
	f, err := os.OpenFile("bb.txt", os.O_CREATE|os.O_WRONLY|os.O_TRUNC, 0600)
	if err != nil {
		panic(err)
	}
	defer f.Close()

	log := slog.New(slog.NewJSONHandler(f, nil))
	b.ResetTimer()
	ctx := context.TODO()
	for n := 0; n < b.N; n++ {
		log.LogAttrs(ctx,
			slog.LevelInfo,
			"hello world",
			slog.Int("index", n),
			slog.String("name", "werben"),
			slog.Int("age", 300),
		)
	}
}

func BenchmarkZerologInfoJson(b *testing.B) {
	f, err := os.OpenFile("cc.txt", os.O_CREATE|os.O_WRONLY|os.O_TRUNC, 0600)
	if err != nil {
		panic(err)
	}
	defer f.Close()

	zlog := zerolog.New(f)
	b.ResetTimer()

	for n := 0; n < b.N; n++ {
		zlog.Info().
			Int("index", n).
			Str("name", "werben").
			Int("age", 300).
			Msg("hello world")
	}
}
```
```
BenchmarkSlogInfo-8          	  313608	      3454 ns/op	       8 B/op	       0 allocs/op
BenchmarkSlogAttr-8          	  323670	      3347 ns/op	       0 B/op	       0 allocs/op
BenchmarkZerologInfo-8       	  126321	      9023 ns/op	    2057 B/op	      59 allocs/op
BenchmarkSlogInfoJson-8      	  337041	      3250 ns/op	       8 B/op	       0 allocs/op
BenchmarkSlogAttrJson-8      	  307329	      3257 ns/op	       0 B/op	       0 allocs/op
BenchmarkZerologInfoJson-8   	  632811	      1779 ns/op	       0 B/op	       0 allocs/op
```

werbenhu:
@dgduncan @thedevop How about this? In this case, if you need to utilize zerolog, you can simply implement your own Logger.
```golang
package logger

import (
	"context"
	"os"

	"golang.org/x/exp/slog"
)

type Level int

const (
	LevelDebug Level = -4
	LevelInfo  Level = 0
	LevelWarn  Level = 4
	LevelError Level = 8
)

var defaultLogger Logger

func init() {
	defaultLogger = NewDefaultLogger()
}

type Logger interface {
	Info(msg string, args ...any)
	Warn(msg string, args ...any)
	Error(msg string, args ...any)
	Debug(msg string, args ...any)
	Fatal(msg string, args ...any)
	Log(level Level, msg string, args ...any)
}

func Default() Logger {
	return defaultLogger
}

type DefaultLogger struct {
	log *slog.Logger
}

func NewDefaultLogger() *DefaultLogger {
	return &DefaultLogger{
		log: slog.New(slog.NewTextHandler(os.Stderr, nil)),
	}
}

func (s *DefaultLogger) Info(msg string, args ...any) {
	s.log.Info(msg, args)
}

func (s *DefaultLogger) Warn(msg string, args ...any) {
	s.log.Warn(msg, args)
}

func (s *DefaultLogger) Error(msg string, args ...any) {
	s.log.Error(msg, args)
}

func (s *DefaultLogger) Fatal(msg string, args ...any) {
	s.log.Error(msg, args)
	os.Exit(1)
}

func (s *DefaultLogger) Debug(msg string, args ...any) {
	s.log.Debug(msg, args)
}

func (s *DefaultLogger) Log(level Level, msg string, args ...any) {
	slevel := slog.Level(level)
	s.log.Log(context.TODO(), slevel, msg, args)
}
```

dgduncan:
@thedevop Thank you for adding those benchmarks. That solution fits along with what we had discussed of in the future creating a logging interface in which you could essentially bring your own logger that slots into the standard interface.

werbenhu:
@dgduncan Transform the logger into an interface for enhanced flexibility. I believe you can proceed with this now while keeping the existing interface consistent. Based on your current groundwork, there will be minimal changes required. Remember to have the server expose a SetLogger capability. Please note that my code was written hastily and is only provided for your reference.

thedevop:
I like the idea of interface, so other logger can be plugged in. One question, given the signature of slog:

werbenhu:
@thedevop  this is indeed a situation where finding a perfect solution is challenging. Different logging libraries do have their own interfaces, and encapsulating them into a single interface might not provide comprehensive coverage. In my opinion, prioritizing the default logging library is a practical approach, as other libraries might require compatibility conversions. However, these conversions do come at a performance cost, and it's possible that we might find ourselves reverting to using a single logging library again. @mochi-co, perhaps it's best for you to make the decision.

dgduncan:
This is where I personally landed on this issue. @werbenhu and @thedevop have come in with some awesome insights and tests. Having a solution such that you can essentially BYOL (bring your own logger) would be an awesome thing to have; however, the lack of a standard `Logger` interface is a problem as old as GO itself and seemingly glaring omission from the standard library. Here are some discussion for those curious: https://github.com/golang/go/issues/28412 and https://github.com/golang/go/issues/13182

The purpose of the move to `slog` was to bring us closer in line with the standard library. Solving the problem of BYOL seems to me like a bit of a pre-optimization for personal configuration and code aesthetic whose layers of complexity are starting to show. All of this to say that I land on the side of sticking with simply using `slog` as implemented and proposing a future feature to work on BYOL.

thedevop:
@dgduncan it's awesome with what you have done with slog. Thanks!

BTW, given this is now using Log.Info, Log.Error, etc... these APIs allow pass in any. So log lines where err.Error() is used can shorten to err. For example:
```go
server.Log.Error("server.InjectPacket", "error", err.Error())
```
to
```go
server.Log.Error("server.InjectPacket", "error", err)
```

server_test.go
```diff
@@ -37,6 +37,11 @@ type AllowHook struct {
 	HookBase
 }
 
+func (h *AllowHook) SetOpts(l *slog.Logger, opts *HookOptions) {
```
werbenhu, line 40: I haven't figured out where the AllowHook.SetOpts() and AllowHook.SetOpts() functions are being used.

werbenhu (commented):
@dgduncan you've done a truly great job.

dgduncan:
I made another update as per the mention of @thedevop of changing`err.Error()` to just `err`

server_test.go
dgduncan: I don't think they currently are in the hooks. They are probably there as placeholder.

dgduncan:
So I have decided to leave the `log.Fatal(err)` as these are technically just examples of how the server can be used and is left up to the user to change the way they want to exit the application.

Unless anyone has any other objections I do believe that this is ready for final review by @mochi-co .

mochi-co:
So much good stuff in this PR, thank you @dgduncan, @werbenhu, @thedevop, @torntrousers. Particularly great to see the benchmarks, helps in making an informed decision. Apologies for being late to the party.

Overall I think this is the right move and I'm excited to get it merged in. My thoughts:

* Where `log.Fatal` has been used, we should keep it as-is. Generally this clearly indicates the end-of-the-road for the runtime and its meaning is unmistakable. 
* Prefer shorthand convenience functions `Info`, `Error`, etc. over `LogAttrs` - the difference in performance is negligible compared to other costs which could be optimised (topics, for example).
* Since the beginning, this project has gone from single embedded logger, to logging interface, back to single embedded logger. In my anecdotal experience, nobody ever reimplemented the built-in logger, so we moved to zerolog directly. I would wager that many packages will migrate towards `log/slog` over the next year to avoid this 'which logger' problem, and as adoption increases, the output format of `log/slog` will become common-place and expected. For this reason, I am inclined to suggest remaining with direct `log/slog` calls for now - if there's a serious demand for a logging interface, we should implement it as a new feature.
* `log/slog` imports should be at the top of the imports block ;) 
* I miss the terminal colours from zerolog.

However, in conclusion I am very excited about this change. @dgduncan you have done a superb job at driving this project forward, and implementing it in a thoughtful and well-considered way. I think this is something you can be proud of.

hooks/storage/badger/badger.go
```diff
@@ -136,7 +137,7 @@ func (h *Hook) OnWillSent(cl *mqtt.Client, pk packets.Packet) {
 // updateClient writes the client data to the store.
 func (h *Hook) updateClient(cl *mqtt.Client) {
 	if h.db == nil {
-		h.Log.Error().Err(storage.ErrDBFileNotOpen)
+		h.Log.Error("", "error", storage.ErrDBFileNotOpen.Error())
```
mochi-co, line 140: I think these `.Error()` calls can removed as @thedevop noted.

mochi-co (changes requested):
I've reviewed this PR, the only recommendation I have is to remove the 48 calls to `errorMsg.Error()` - as @thedevop noted, this is no longer needed when using `log/slog`. Otherwise, looks good to me!

hooks/storage/badger/badger.go
dgduncan: Good catch.

dgduncan: Pushed changes to fix all references.

mochi-co (approved):
Approved

werbenhu (approved):

thedevop:
@dgduncan , as I prepare my application for this break change (as I do override default logger), I ran into few issues, maybe you have some insights to simple solutions:
1. slog with JSONHandler can't log struct unless it implements LogValuer. I can do this for my own struct, but can't really do that for 3rd party struct. Using TextHandler works as struct->string is built-in.
2. if the object is already in json, slog can't just add the raw json without adding all the escape character to treat it as one long string.

dgduncan:
@thedevop Would you mind explaining providing an example of a struct that is failing to be logged? I have not run into this issue yet in my own use. Do you mean that there is something `struct` that is attempting to be logged and there is something inside said `struct` that fails which breaks the record or shows a `BadKey` because it does not implement either interface?

mochi-co:
I want to see what @thedevop says before I merge this in case there's a compelling reason that we really should be adding an interface.

dgduncan:
@mochi-co I am curious as well. I do think that if this is a problem in only very specific use cases you in theory should be able to wrap the struct in another struct that implements `LogValue` and when reflect over that struct that is causing problems if you wanted to mirror that struct to log or build the `json.RawMessage` map by manually going over that struct that is causing the problem.

The thing I do want to know is if this is a problem for structs that are being logged in the broker code itself or if this is a problem with code from in the embedding application that will now have to use `slog` with this change. Because the first is breaking the broker itself and the other is a implementation challenge of moving to a new standard.

thedevop:
This is a problem for any struct that doesn't implement LogValuer (or json. Marshaler) interface AND using JSONHandler.

Here is an example that demonstrates the issue: https://go.dev/play/p/KNsC_CmrmSG. I also have B which implements (commented out).

werbenhu:
@thedevop It's strange, but I can run it without any issues using golang.org/x/exp/slog. Take a look: [https://go.dev/play/p/Q48CYRqnPIb](https://go.dev/play/p/Q48CYRqnPIb)

werbenhu:
@thedevop Just making Name and Value of struct A exported is fine. [https://go.dev/play/p/8H4doeulXIS](https://go.dev/play/p/8H4doeulXIS)
```
package main

import (
	"log/slog"
	"os"
)

type A struct {
	Name  string
	Value int
}

func main() {
	textLog := slog.New(slog.NewTextHandler(os.Stdout, nil))
	jsonLog := slog.New(slog.NewJSONHandler(os.Stdout, nil))
	a := A{"test", 123}
	textLog.Info("text without logvalue", "a", a)
	jsonLog.Info("json without logvalue", "a", a)
}
```

werbenhu:
@thedevop Just making Name and Value of struct A exported is fine. [https://go.dev/play/p/8H4doeulXIS](https://go.dev/play/p/8H4doeulXIS)
```
package main

import (
	"log/slog"
	"os"
)

type A struct {
	Name  string
	Value int
}

func main() {
	textLog := slog.New(slog.NewTextHandler(os.Stdout, nil))
	jsonLog := slog.New(slog.NewJSONHandler(os.Stdout, nil))
	a := A{"test", 123}
	textLog.Info("text without logvalue", "a", a)
	jsonLog.Info("json without logvalue", "a", a)
}
```

thedevop:
@werbenhu , you're right! I missed that because the TextHandler was able log. Thanks!

mochi-co:
If everyone is happy, I will merge this branch. Please let me know 🙂

dgduncan:
> What Hath God Wrought. - Samuel Morse

mochi-co:
Merged! Amazing, immense work by everyone, especially @dgduncan! 🥳 
This will probably go out between now and the weekend as a new minor version 2.4.0 🙂

PR #252, opened 2023-07-10
WriterSize parameter is incorrectly set

wind-c:
The WriterSize parameter is incorrectly set in the newClient method.


mochi-co (approved):
Good catch!

PR #253, opened 2023-07-10
Ensure msg doesn't exceed subscription QoS

thedevop:
If subscriber QoS is less than the publisher QoS, downgrade delivery QoS for that subscriber.

In addition, the below code can be removed, as the during the subscription, we already check if it exceed MaxmumQos:
```go
	if out.FixedHeader.Qos > s.Options.Capabilities.MaximumQos {
		out.FixedHeader.Qos = s.Options.Capabilities.MaximumQos // [MQTT-3.2.2-9]
	}
```

### DIFF
--- a/server.go
+++ b/server.go
@@ -818,6 +818,10 @@ func (s *Server) publishToClient(cl *Client, sub packets.Subscription, pk packet
 		sort.Ints(out.Properties.SubscriptionIdentifier)
 	}
 
+	if out.FixedHeader.Qos > sub.Qos {
+		out.FixedHeader.Qos = sub.Qos
+	}
+
 	if out.FixedHeader.Qos > s.Options.Capabilities.MaximumQos {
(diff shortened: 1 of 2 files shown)


mochi-co:
> In addition, the below code can be removed, as the during the subscription, we already check if it exceed MaxmumQos:
> if out.FixedHeader.Qos > s.Options.Capabilities.MaximumQos {
>		out.FixedHeader.Qos = s.Options.Capabilities.MaximumQos // [MQTT-3.2.2-9]
>	}

We check it in processPublish (when a publish is received via normal channels), and via publishToClient. Because processPublish indirectly calls publishToClient, we can remove it from processPublish. We can't remove it from publishToClient as not all publish packets are injected via processPublish. I'll remove it in the next release 👍🏻 

Otherwise looks good to me, nice work!

mochi-co (approved):
An excellent contribution

Issue #249, opened 2023-06-27
Why does Publish call OnPublish ?


Libong:
What is the meaning of this code? If I call the Publish or InjectPacket function, will the content sent by the server be passed back to OnPublish for the server to receive

if pkx, err := s.hooks.OnPublish(cl, pk); err == nil {
		pk = pkx
	} else if errors.Is(err, packets.ErrRejectPacket) {
		return nil
	}

thedevop:
Yes, OnPublish will be called. You can filter out in the hook by checking for `cl.Net.Inline`.

Libong:
I see in the code that only when the Publish function that encapsulates InjectPacket is issued, will the inline be set to true. If I call InjectPacket instead of Publish, how can I filter it? Alternatively, you can only call the Publish function

thedevop:
Publish is merely a convenient function that wrap InjectPacket.
https://github.com/mochi-co/mqtt/blob/0648e395075ee3072747d259e777b7166619217c/server.go#L647-L649

If you directly use InjectPacket, you should also use an inline client.

To filter out inline publish in hook:
```go
	if cl.Net.Inline {
		return pk, nil
	}
```

oherych:
Hi. I have additional questions about this code. Why do we ignore errors? What should be in case if we have have error different from `packets.ErrRejectPacket`. For example in my case I want to send `packets.ErrPayloadFormatInvalid`

```go
	if pkx, err := s.hooks.OnPublish(cl, pk); err == nil {
		pk = pkx
	} else if errors.Is(err, packets.ErrRejectPacket) {
		return nil
	}
```

mochi-co:
@oherych `packets.ErrRejectPacket` is used in order to silently drop/reject an incoming packet. 

The [MQTTv5 Spec](https://docs.oasis-open.org/mqtt/mqtt/v5.0/os/mqtt-v5.0-os.html#_Toc3901120) doesn't detail a standardised mechanism for returning error messages to clients. That said - QoS > 0 packets do return Puback/Pubrec packets on receipt of a QoS packet, and in MQTTv5 both the reason code and reason string may be set in response.

In your case, you could use `cl.WritePacket` to issue a an ack with the packets.ErrPayloadFormatInvalid error code and reason string the client, and then return a `packets.ErrRejectPacket` as the err response for `s.hooks.OnPublish`. For example, something like this:

```go
packets.Packet{
	FixedHeader: packets.FixedHeader{
        Type: packets.Puback,
[...]

PR #256, opened 2023-07-13
Allow Publish to return custom Ack error responses

mochi-co:
Following a discussion in #249, it became apparent that we do not currently return packet errors in acks when a received publish packet is invalid (or an error has been returned by the OnPublish hook), which is a behaviour indicated in the MQTT v5 specification. 

This PR adds a check statement to return an error ack if the client is v5, the error is a packet error (not a 'true error'), and the Qos > 0. 

Additionally, a line has been corrected allowing the User Properties to be propagated to the ack, as per the spec (formerly this would be dropped). As such, a compatibility mode has been added to bypass a spec violation in the paho interoperability library which expects no user properties. We should consider migrating away from the paho testing library.

Otherwise passing Paho and Unit tests.

I can't seem to add anyone for review, so @thedevop @wind-c @dgduncan, if you could give this a look over, I would be grateful 🙏🏻

### DIFF
--- a/server.go
+++ b/server.go
@@ -71,10 +71,11 @@ type Capabilities struct {
 
 // Compatibilities provides flags for using compatibility modes.
 type Compatibilities struct {
-	ObscureNotAuthorized     bool // return unspecified errors instead of not authorized
-	PassiveClientDisconnect  bool // don't disconnect the client forcefully after sending disconnect packet (paho)
-	AlwaysReturnResponseInfo bool // always return response info (useful for testing)
-	RestoreSysInfoOnRestart  bool // restore system info from store as if server never stopped
+	ObscureNotAuthorized       bool // return unspecified errors instead of not authorized
+	PassiveClientDisconnect    bool // don't disconnect the client forcefully after sending disconnect packet (paho - spec violation)
+	AlwaysReturnResponseInfo   bool // always return response info (useful for testing)
+	RestoreSysInfoOnRestart    bool // restore system info from store as if server never stopped
+	NoInheritedPropertiesOnAck bool // don't allow inherited user properties on ack (paho - spec violation)
 }
 
 // Options contains configurable options for the server.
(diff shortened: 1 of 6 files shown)


server.go
```diff
@@ -715,10 +716,14 @@ func (s *Server) processPublish(cl *Client, pk packets.Packet) error {
 		pk.FixedHeader.Qos = s.Options.Capabilities.MaximumQos // [MQTT-3.2.2-9] Reduce Qos based on server max qos capability
 	}
 
-	if pkx, err := s.hooks.OnPublish(cl, pk); err == nil {
+	pkx, err := s.hooks.OnPublish(cl, pk)
+	if err == nil {
 		pk = pkx
 	} else if errors.Is(err, packets.ErrRejectPacket) {
```
thedevop, line 722: Can we add a new internal Error code (Ignore) here. Basically it will not further publish to the subscribers, but it will build the proper acks. Currently, ErrRejectPacket works like that only for QoS 0.

packets.ErrRejectPacket can be captured in the next else if.

server.go
```diff
@@ -715,10 +716,14 @@ func (s *Server) processPublish(cl *Client, pk packets.Packet) error {
...
 		return nil
+	} else if err != nil && errors.As(err, new(packets.Code)) && cl.Properties.ProtocolVersion == 5 && pk.FixedHeader.Qos > 0 {
+		_ = cl.WritePacket(s.buildAck(pk.PacketID, packets.Puback, 0, pk.Properties, err.(packets.Code))) // omit connection error in favour of existing packet err
```
thedevop, line 725: When return err on the next line, it will result in disconnect with the reason as well.

mochi-co: This was intentional as we already disconnect the client on any invalid packet due to protocol violation - but now I'm wondering if it's the correct behaviour for acks. What do you think?

server.go
mochi-co: > Currently, ErrRejectPacket works like that only for QoS 0.

I'm not sure this is the case because
```go
} else if errors.Is(err, packets.ErrRejectPacket) {
    return nil
```
will return before `s.publishToSubscribers(pk)` occurs 10 lines later. As it stands, RejectPacket will effectively behave if the packet never existed.

We could add a `packets.ErrIgnorePacket` as you say, which will simply not retain or publish out the incoming message, but pretend that the packet was received normally. Is this what you mean?

thedevop: Yes. Ignore will pretend that packet is received normally, but will not retain or publish.

server.go
thedevop: Looking at the spec, only the following reason code can be return on PUBACK:
https://docs.oasis-open.org/mqtt/mqtt/v5.0/os/mqtt-v5.0-os.html#_Toc3901124

I would think for these reason codes, it should not result in a disconnect for V5.

mochi-co:
I'm feeling pretty good about this so I'm going to merge it

Issue #251, opened 2023-07-09
Incorrect(?) handling of 'retain' flag in MQTT 3.1.1


bkupidura:
Looks like `mochi-co/mqtt` is incorrectly setting `retain` flag for published messages in 3.1.1.

Based on `MQTT-3.3.1-9` 
```
It MUST set the RETAIN flag to 0 when a PUBLISH Packet is sent to a Client because it matches an established subscription regardless of how the flag was set in the message it received [MQTT-3.3.1-9].
```

So imho broker after receiving retained message, should store it - but packet should have `retain` flag set to `0` when publishing to clients, unless packet is really retained (and not live sent). 

How mosquitto is handling that:

Scenario 1:
1) Client-A subscribes to `test/test`
2) Client-B sends retained message to `test/test`
3) Client-A receives message, but retained flag is `0`

Scenario 2:
1) Client-B sends retained message to `test/test`
2) Client-A subscribes to `test/test`
3) Client-A receives message, retained flag set to `1`

Currently `mochi-co/mqtt` is always setting retained flag to `1` in publish if packet was send with retained flag set (in mqtt 3.1.1).

https://github.com/mochi-co/mqtt/blob/master/server.go#L809

mochi-co:
This should be resolved in [v2.2.16](https://github.com/mochi-co/mqtt/releases/tag/v2.2.16), thanks again :)

PR #257, opened 2023-07-13
Retain flag should be delivered as false in v3

mochi-co:
Per @bkupidura's note in #251, the retain flag was incorrectly being transmitted as set to MQTTv3 clients. For MQTTv3, the retain flag should be delivered as 'false' after being stored.

This PR adds a correction to this behaviour, and adjusts any tests implementing v3 retain accordingly

cc @bkupidura @thedevop @dgduncan @wind-c - if anyone wishes to review :)

### DIFF
--- a/packets/packets.go
+++ b/packets/packets.go
@@ -175,6 +175,7 @@ type Subscription struct {
 	Qos               byte
 	RetainAsPublished bool
 	NoLocal           bool
+	FwdRetainedFlag   bool // true if the subscription forms part of a publish response to a client subscription and packet is retained.
 }
 
 // Copy creates a new instance of a packet, but with an empty header for inheriting new QoS flags, etc.
@@ -387,7 +388,7 @@ func (pk *Packet) ConnectDecode(buf []byte) error {
 		offset += n
 	}
 
-	pk.Connect.ClientIdentifier, offset, err = decodeString(buf, offset) //[MQTT-3.1.3-1] [MQTT-3.1.3-2] [MQTT-3.1.3-3] [MQTT-3.1.3-4]
+	pk.Connect.ClientIdentifier, offset, err = decodeString(buf, offset) // [MQTT-3.1.3-1] [MQTT-3.1.3-2] [MQTT-3.1.3-3] [MQTT-3.1.3-4]
 	if err != nil {
 		return ErrClientIdentifierNotValid // [MQTT-3.1.3-8]
 	}
--- a/server.go
+++ b/server.go
@@ -806,7 +806,7 @@ func (s *Server) publishToClient(cl *Client, sub packets.Subscription, pk packet
 	}
 
 	out := pk.Copy(false)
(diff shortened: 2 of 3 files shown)


bkupidura:
@mochi-co didnt had time yet to deploy this change and verify. But AFAIK with this PR you will always set `retain` flag to `0` for MQTT < 5.

`retain` flag should be set to `0` if msg is delivered to already subscribed customers, but if customer if subscribed *AFTER* publish - `retain` should be set to `1`.
At least this is how mosquitto implemented that. Tomorrow morning i will try to find reason in RFC

mochi-co:
> retain flag should be set to 0 if msg is delivered to already subscribed customers, but if customer if subscribed AFTER publish - retain should be set to 1.

Ahhhh. It's been a while since I read the v3 spec. Let me know and I'll have a think in the meantime. Hopefully we can get it into this weekend's release.

bkupidura:
Ok, got it:
```
When sending a PUBLISH Packet to a Client the Server MUST set the RETAIN flag to 1 if a message is sent as a result of a new subscription being made by a Client [MQTT-3.3.1-8]. It MUST set the RETAIN flag to 0 when a PUBLISH Packet is sent to a Client because it matches an established subscription regardless of how the flag was set in the message it received [MQTT-3.3.1-9]. 
```

mochi-co:
@bkupidura Perfect, let's think if there is a simple way to make this change 👍🏻

thedevop:
After reading the spec, for both V5 and V3, it seems the retain flag should be set only when a client subscribes to a topic (not including inheriting a subscription through a previous session). If so, then publishRetainedToClient has that knowledge, but it doesn't pass on info to publishToClient. Perhaps we can add a new parameter to publishToClient to pass in a retain flag?

mochi-co:
I was not entirely happy with the idea of adding more parameters to publishToClient, however I added a new parameter to `packets.Subscription` called `FwdRetainedFlag` which is set to true when the subscription used is the result of publishing retained messages from a subscribe action. However, double check if this makes sense.

bkupidura:
LGTM

```
# mosquitto_sub -t test -v -V mqttv311 -F 'retained:%r topic:%t data:%p'
retained:1 topic:test data:published-before-sub [MQTT-3.3.1-8]
retained:0 topic:test data:published-after-sub [MQTT-3.3.1-9]

# mosquitto_sub -t test -v -V mqttv5 -F 'retained:%r topic:%t data:%p'
retained:1 topic:test data:published-before-sub [MQTT-3.3.1-12]
retained:0 topic:test data:published-after-sub [MQTT-3.3.1-12]

# mosquitto_sub -t test -v -V mqttv5 --retain-as-published -F 'retained:%r topic:%t data:%p'
retained:1 topic:test data:published-before-sub [MQTT-3.3.1-13]
retained:1 topic:test data:published-after-sub [MQTT-3.3.1-13]
retained:0 topic:test data:published-after-sub-without-retain [MQTT-3.3.1-13]
```

thedevop:
That works.

PR #260, opened 2023-07-14
Fix websocket reads for packets > 1 buffer size

ianrose14:
Fixes https://github.com/mochi-co/mqtt/issues/259

### DIFF
--- a/listeners/websocket.go
+++ b/listeners/websocket.go
@@ -99,7 +99,7 @@ func (l *Websocket) handler(w http.ResponseWriter, r *http.Request) {
 	}
 	defer c.Close()
 
-	err = l.establish(l.id, &wsConn{c.UnderlyingConn(), c})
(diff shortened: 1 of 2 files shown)


mochi-co (approved):
Thanks @ianrose14!

This looks good to me. As far as I can tell it's correct, and it works as far as I can see in my tests and websocket connections. I approve 👍🏻 If nobody has any concerns, I will probably merge it in tomorrow.

ianrose14:
> Thanks @ianrose14!
> 
> This looks good to me. As far as I can tell it's correct, and it works as far as I can see in my tests and websocket connections. I approve 👍🏻 If nobody has any concerns, I will probably merge it in tomorrow.

Great, thanks for the fast review!

Also thanks for providing this package - we were previously using Google Cloud's IoT service, but that's getting shut down next month (😭) and this package has been really nice for developing our in-house alternative.   👏

Issue #258, opened 2023-07-14
ACLs that allow only certain topics


heya-naohiro:
Hello, This is my first Issue on this project.

If I want to allow only certain topics and exclude others, can I use "github.com/mochi-co/mqtt/v2/hooks/auth" to implement this?
The following implementation seems to work at first glance, but fails one out of several times.
The following will be output as a debug message 
```DBG client failed allowed ACL check```

```
package main

import (
	"log"
	"os"
	"os/signal"
	"syscall"

	"github.com/mochi-co/mqtt/v2"
	"github.com/mochi-co/mqtt/v2/hooks/auth"
	"github.com/mochi-co/mqtt/v2/listeners"
	"github.com/rs/zerolog"
)

func main() {
	server := mqtt.New(nil)
	l := server.Log.Level(zerolog.DebugLevel)
	server.Log = &l

	rules := &auth.Ledger{
		Auth: auth.AuthRules{
			{Remote: "*", Allow: true},
		},
		ACL: auth.ACLRules{ 
			{
				Filters: auth.Filters{
					"sometopic": auth.ReadWrite,
					"#":         auth.Deny,
				},
			},
		},
	}

	err := server.AddHook(new(auth.Hook), &auth.Options{
		Ledger: rules,
	})
	if err != nil {
		log.Fatal(err)
	}

	tcp := listeners.NewTCP("t1", ":1885", nil)

	err = server.AddListener(tcp)
	if err != nil {
		log.Fatal(err)
	}

	sigs := make(chan os.Signal, 1)
	done := make(chan bool, 1)
	signal.Notify(sigs, syscall.SIGINT, syscall.SIGTERM)
	go func() {
		<-sigs
		done <- true
	}()

	go func() {
		err = server.Serve()
		if err != nil {
			log.Fatal(err)
		}
	}()
	<-done
	server.Close()
}

```

# MQTT-Server Log

```
6:28AM INF added hook hook=auth-ledger
6:28AM INF loaded auth rules acl=1 authentication=1 hook=auth-ledger
6:28AM INF attached listener address=:1885 id=t1 protocol=tcp
6:28AM INF mochi mqtt starting version=2.2.15
6:28AM INF mochi mqtt server started
6:28AM DBG system event loop started
6:28AM DBG client disconnected client=mosq-G2QKl8k6D4DGn1Rml3 listener=t1 remote=127.0.0.1:52012
[...]

PR #262, opened 2023-07-14
Preference Write, Read, Deny filters in ledger

mochi-co:
Fixes issue #258 in which the filters for the ledger ACL rules were selected at random due to the behaviour of the `range` operator. This PR changes the order of operations to first check for Write access filters, then Read filters, and finally Deny filters.

It's slightly less optimised than it could be, but it should be fixed first, and optimised later if anyone feels the need to do so.

### DIFF
--- a/hooks/auth/ledger.go
+++ b/hooks/auth/ledger.go
@@ -9,9 +9,10 @@ import (
 	"strings"
 	"sync"
 
+	"gopkg.in/yaml.v3"
+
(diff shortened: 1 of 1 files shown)


dgduncan (approved):
I think it looks great! I have a vague memory of a more optimized solution for weighted ruling similar to this; however, the solution escapes me. Great slightly more common than edge case find!

PR #264, opened 2023-07-19
method UnsubscribeClient's packet add fixedHeader

kenuestar:
When calling the OnUnsubscribed method during secondary development, the corresponding processor will be selected based on the type of the packet. However, the lack of packet type in UnsubscribeClient results in the inability to select the appropriate processor successfully.

### DIFF
--- a/server.go
+++ b/server.go
@@ -1144,7 +1144,7 @@ func (s *Server) UnsubscribeClient(cl *Client) {
 		filters[i] = v
 		i++
 	}
-	s.hooks.OnUnsubscribed(cl, packets.Packet{Filters: filters})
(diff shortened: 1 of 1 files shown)


server.go
```diff
@@ -1144,7 +1144,7 @@ func (s *Server) UnsubscribeClient(cl *Client) {
 		filters[i] = v
 		i++
 	}
-	s.hooks.OnUnsubscribed(cl, packets.Packet{Filters: filters})
+	s.hooks.OnUnsubscribed(cl, packets.Packet{FixedHeader: packets.FixedHeader{Type: packets.Unsubscribe}, Filters: filters})
```
mochi-co, line 1147: This feels unusual as there's an implicit assumption that any packet going to `OnUnsubscribed` would be an unsubscribe packet. However, I don't see anything fundamentally wrong with this, and it does make some sense. So I'm approving it 👍🏻

mochi-co (approved):

PR #266, opened 2023-07-20
fix: data-race in badger hook

gsagula:
Potential fix for https://github.com/mochi-co/mqtt/issues/265


dgduncan:
I will let others speak on this  as I am not super familiar with the storage implementation, but them seems to be a safe way to prevent said race conditions.

mochi-co:
@gsagula This looks like it would work, as far as I can see. With the change in place, do you find it fixes any race conditions you are seeing?

mochi-co (approved):
I've tested this and confirm it fixes the issue - nice work @gsagula!

PR #267, opened 2023-07-21
Use JSONeq to compare JSON

wind-c:
This ensures that test results pass even if the field order is inconsistent.


mochi-co (approved):
Great idea, I was always wondering if something like this existed but never thought to look. Thank you!

PR #269, opened 2023-07-24
fix issue: incorrect subscribe scene of share subscribe.

x20080406:
scene:
cl1 subscribe with filters: $SHARE/g1/a/b/c; a/b/c
cl2 subscribe with filter: $SHARE/g1/a/b/c
cl3 publish a message with topic a/b/c through broker. 

actully: cl1 always receive message. but sometimes cl2 receive message. 
correct: cl2 never receive message. cl1 receive message.

I have a test on emqx. it's correct, cl2 never receve message.

@mochi-co please review.

### DIFF
--- a/topics.go
+++ b/topics.go
@@ -251,10 +251,16 @@ type Subscribers struct {
 	Subscriptions  map[string]packets.Subscription
 }
 
-// SelectShared returns one subscriber for each shared subscription group.
 func (s *Subscribers) SelectShared() {
 	s.SharedSelected = map[string]packets.Subscription{}
+p1:
 	for _, subs := range s.Shared {
+		for client, _ := range subs {
+			if _, ok := s.Subscriptions[client]; ok {
+				continue p1
+			}
+		}
+
 		for client, sub := range subs {
 			cls, ok := s.SharedSelected[client]
 			if !ok {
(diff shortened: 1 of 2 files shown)


thedevop:
@x20080406, looking through the MQTT spec, the currently implementation should be the correct behavior:
* It states `Each Shared Subscription is independent from any other. It is possible to have two Shared Subscriptions with overlapping filters. In such cases a message that matches both Shared Subscriptions will be processed separately by both of them. If a Client has a Shared Subscription and a Non‑shared Subscription and a message matches both of them, the Client will receive a copy of the message by virtue of it having the Non‑shared Subscription. A second copy of the message will be delivered to one of the subscribers to the Shared Subscription, and this could result in a second copy being sent to this Client.`
* It also mentions `If there's more than one Session subscribed to the Shared Subscription, the Server implementation is free to choose, on a message by message basis, which Session to use and what criteria it uses to make this selection.`

x20080406:
> * Each Shared Subscription is independent from any other

sure, I've got it.

mochi-co:
Just to follow up on this @x20080406 - have a look at the hook `OnSelectSubscribers`, this hook allows you to modify the list of subscribers being published to, including the selected subscriber from the shared subscription list. 
The idea is you might have your hook check against some other dataset to determine (like client load, or a priority list) which is the preferred client to send to in a shared sub scenario.

PR #276, opened 2023-08-08
Fix readme example

dgduncan:
Update the example in the README to better reflect real world use of the server. This includes adding signals, creating a go routine for `server.Serve()` and blocking until interrupted.

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,6 +94,16 @@ import (
 )
 
 func main() {
+  
+  // Create signals channel to run server until interrupted
+  sigs := make(chan os.Signal, 1)
+	done := make(chan bool, 1)
+	signal.Notify(sigs, syscall.SIGINT, syscall.SIGTERM)
+	go func() {
+		<-sigs
+		done <- true
+	}()
+
   // Create the new MQTT Server.
   server := mqtt.New(nil)
   
@@ -107,10 +117,18 @@ func main() {
     log.Fatal(err)
   }
   
-  err = server.Serve()
-  if err != nil {
-    log.Fatal(err)
-  }
+
+  go func() {
+		err := server.Serve()
+		if err != nil {
+			log.Fatal(err)
+		}
+	}()
+
+  // Run server until interrupted
+  <-done
+
+  // Cleanup
 }
 ```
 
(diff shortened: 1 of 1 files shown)


mochi-co (dismissed):
This is a great idea, it more accurately represents real world usage and will help people get up and running easier. Approved! :)

README.md
```diff
@@ -103,10 +112,18 @@ func main() {
     log.Fatal(err)
   }
   
-  err = server.Serve()
-  if err != nil {
-    log.Fatal(err)
-  }
+
+  go func() {
```
torntrousers, line 116: The server.Serve() doesn't actually need to be Go routine, does it? I suppose it doesn't matter much either way, but for the sake of keeping the readme example simple?

README.md
```diff
@@ -90,6 +90,15 @@ import (
...
+  sigs := make(chan os.Signal, 1)
+  done := make(chan bool, 1)
+  signal.Notify(sigs, syscall.SIGINT, syscall.SIGTERM)
+  go func() {
+    <-sigs
+    done <- true
+  }()
+
   // Create the new MQTT Server.
   server := mqtt.New(nil)
```
torntrousers, line 103: To get this to work I had to change the import to be ```mqtt "github.com/mochi-mqtt/server/v2"```

README.md
dgduncan: It technically does not; however, in many of code [examples](https://github.com/mochi-mqtt/server/tree/fix-readme-example/examples) it is done with a go routine so I did this to remain consistent.

README.md
dgduncan: Very good catch! I had to do the same thing but I forgot to copy over imports.

README.md
dgduncan: I have no strong opinion either way. Perhaps we could wait to hear from @mochi-co to see if there was an original intent behind using the new go routine? I could see a world in which that was left over from a time when `.Serve()` did block which made it more necessary for the new routine.

torntrousers (approved):
That change to all the examples could be another PR though and this one LGTM

mochi-co (approved):

PR #283, opened 2023-08-26
Inline Client Subscriptions prototype

mochi-co:
Following the discussion in #279 (_Direct subscribe for embedded use case_), I've put together a prototype implementation of inline subscriptions as a counterpart of the `server.Publish` method.

I thought about this for almost two weeks and implemented several different versions. There were three core solutions which met the requirements of the direct subscription feature:

- **Hook Model**: An implementation which used the existing hook system and added an additional hook beside `OnPublished`. The user could add new 'subscriptions' - topic filters and callback methods, and then the hook would be triggered when a message is published. Any subscriptions on this hook would be iterated, and a new internal topic comparison method (i.e. does topic name match filter) would be used to determine if the message should be provided to the given callback.  
  - Good: Fairly simple to implement. Purely observational approach - didn't meaningfully interact with any existing internal code.
  - Not so good: didn't meaningfully interact with any existing internal code, so most topic features were unusable (e.g. retained), not reflective of actual usage. Required implementation of an entirely new topic matching system, which is slow at scale.
- **Outbound channel**: We already implement an `outbound` channel which receives messages that will be sent to the client. I had some ideas around exposing this channel for developers to access.
  - Good: Using channels would have felt 'proper' and 'sophisticated'.
  - Not so good: All subscribed messages would come through on the same source. Still requires topic matching on the developer side. No clear way to make the subscriptions.
- **Inline Client changes**: We already use special 'inline' clients in order to publish messages directly (via `server.Publish` and `server.InjectPacket`. It seemed reasonable that we could also allow inline clients to have subscriptions too, which triggered callbacks instead of writing packets to the network.
   -  Good: Works almost exactly as if the client was external. Can receive retained messages. Simple to understand.
   - Not so good: Weird things might happen if you injected a totally custom subscription packet via `server.InjectPacket`. However, this customisability may also be a benefit to some. More Inline Client specific code.

Ultimately, the inline client approach was the only implementation which made both technical and intuitive sense. 

There are two new methods:

- `func (s *Server) Subscribe(cl *Client, filter string, id int, callback InlineSubFn) error`, and
- `func (s *Server) Unsubscribe(cl *Client, filter string) error`

`id` is an MQTTv5 subscription identifier. These methods are used like so:

```go
callbackFn := func(cl *mqtt.Client, sub packets.Subscription, pk packets.Packet) {
  // do something ...
}
err = server.Subscribe(cl, "direct/#", 2, callbackFn)
```

and

```go
err := server.Unsubscribe(cl, "direct/#")
```

I have added a new example file, `examples/direct/main.go` which demonstrates the behaviour, and when run, outputs the following:

```
 % go run examples/direct/main.go
11:17PM INF added hook hook=allow-all-auth
11:17PM INF mochi mqtt starting version=2.3.0
11:17PM INF inline client subscribing
11:17PM INF inline client received message from subscription payload="some other retained message" topic=direct/alternate/retained
11:17PM INF mochi mqtt server started
11:17PM INF inline client received message from subscription payload="retained message" topic=direct/retained
11:17PM INF main.go issued direct message to direct/publish
11:17PM INF inline client received message from subscription payload="scheduled message" topic=direct/publish
11:17PM INF main.go issued direct message to direct/publish
11:17PM INF inline client received message from subscription payload="scheduled message" topic=direct/publish
11:17PM INF main.go issued direct message to direct/publish
11:17PM INF inline client received message from subscription payload="scheduled message" topic=direct/publish
11:17PM INF inline client unsubscribing
11:17PM INF main.go issued direct message to direct/publish
11:17PM INF main.go issued direct message to direct/publish
^C11:17PM WRN caught signal, stopping...
11:17PM INF stopping hook hook=allow-all-auth
11:17PM INF mochi mqtt server stopped
11:17PM INF main.go finished
```

There are no tests attached to this PR, but will be added upon acceptance. It also contains a small commit fixing some typos. Feedback encouraged!

### DIFF
--- a/clients.go
+++ b/clients.go
@@ -107,12 +107,18 @@ type Client struct {
 
 // ClientConnection contains the connection transport and metadata for the client.
 type ClientConnection struct {
-	Conn     net.Conn          // the net.Conn used to establish the connection
-	bconn    *bufio.ReadWriter // a buffered net.Conn for reading packets
-	Remote   string            // the remote address of the client
-	Listener string            // listener id of the client
-	Inline   bool              // client is an inline programmetic client
+	Conn       net.Conn               // the net.Conn used to establish the connection
+	bconn      *bufio.ReadWriter      // a buffered net.Conn for reading packets
+	Remote     string                 // the remote address of the client
+	Listener   string                 // listener id of the client
+	Inline     bool                   // client is an inline programmatic client
+	InlineSubs map[string]InlineSubFn // a map of inline subs callback functions keyed on subscription id
 }
+
+// InlineSubFn is the signature for a callback function which will be called
+// when an inline client receives a message on a topic it is subscribed to.
+// The sub argument contains information about the subscription that was matched for any filters.
+type InlineSubFn func(cl *Client, sub packets.Subscription, pk packets.Packet)
 
 // ClientProperties contains the properties which define the client behaviour.
 type ClientProperties struct {
--- a/clients.go
+++ b/clients.go
@@ -99,7 +99,7 @@ func (cl *Clients) GetByListener(id string) []*Client {
 type Client struct {
 	Properties   ClientProperties // client properties
 	State        ClientState      // the operational state of the client.
-	Net          ClientConnection // network connection state of the clinet
+	Net          ClientConnection // network connection state of the client
 	ID           string           // the client id.
 	ops          *ops             // ops provides a reference to server ops.
 	sync.RWMutex                  // mutex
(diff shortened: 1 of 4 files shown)


dgduncan:
I'll take a look at this in the next few days. I think you went with the right approach with the inline client. I think it makes the most sense most technically and for DX.

bkupidura-JNPR:
LGTM, but ... :)
Is there any reason why `InlineSubs` are part of `ClientConnection` and not `ClientState`?
Current implementation leads to some inconsistency, regular clients subs are part of `ClientState`, inline subs are part of `ClientConnection`.
I would say that having all subs in one struct would be more elegant and clean.

bkupidura:
LGTM, but ... :)
Is there any reason why InlineSubs are part of ClientConnection and not ClientState?
Current implementation leads to some inconsistency, regular clients subs are part of ClientState, inline subs are part of ClientConnection.
I would say that having all subs in one struct would be more elegant and clean.

mochi-co:
Withdrawn in favour of #284

PR #284, opened 2023-08-28
Another code implementation for Inline Client Subscriptions.

werbenhu:
This is another implementation I wrote for Inline Client Subscriptions. It was written in a bit of a rush, and I didn't include test cases. I'm sharing it here for everyone to take a look as discussions are ongoing.


mochi-co:
I've only had a brief look because I'm not at the computer but so far this looks nicer than my implementation 👍🏻

werbenhu:
I'll take some time later to adjust the code and add test cases.

dgduncan:
I will say I do like this implementation quite a bit. By creating a first class inline subscription it makes it easy to follow the breadcrumbs and know how things are interacting. 

@mochi-co I see last night you posted [this](https://github.com/orgs/mochi-mqtt/discussions/279#discussioncomment-6838749). How are you feeling now about creating a inline subscribe? Or simply handing it off to a hooks implementation?

mochi-co:
@werbenhu I've reviewed this code and it's excellent, in terms of both technical quality and overall vision. I am especially happy with the number of tests, the way they've been thought through, and the coverage. I don't really have any change requests. In terms of achieving a specific goal, this PR does it well. I have a few ideas for how we could extend it to add more functionality, but we can discuss.

I have a thought I'd like us to consider before we can move forward, which I've added in #279 - I'd like both of your insight on it if you have a moment to do so.

werbenhu:
@mochi-co @dgduncan I have completed the implementation for inline subscriptions with retained messages and have also added support for multiple callbacks per topic using different inline client IDs.

werbenhu:
I didn't use newClient() here to create an InlineClient for subscriptions. I found that omitting the client makes the logic seem simpler and more straightforward.

werbenhu:
@mochi-co I have a question now, which is whether inline subscriptions should trigger hook.OnSubscribe. If necessary, it might be more appropriate to use a client struct as a parameter in the server.Subscribe() function. If that's the case, this client struct can only be stored within the InlineSubscription structure.

packets/packets.go
```diff
@@ -181,8 +181,9 @@ type Subscription struct {
 
 // InlineSubscription represents an internal subscription, and the Handler is a callback function used to process messages.
 type InlineSubscription struct {
-	Filter  string
-	Handler func(filter string, packet Packet)
+	Filter     string
+	Identifier string
```
mochi-co, line 185: This can be confused with Subscription Identifiers, which are integers. I recommended renaming to `ClientID string` to be less specific. I did consider just `ID`, but this is still ambiguous in some of the later calls and params.

```suggestion
	ClientID string
```

server.go
```diff
@@ -661,23 +661,43 @@ func (s *Server) Publish(topic string, payload []byte, retain bool, qos byte) er
...
+	if handler == nil {
+		return packets.ErrInlineSubscriptionHandlerInvalid
+	} else if !IsValidFilter(filter, false) {
+		return packets.ErrTopicFilterInvalid
+	} else if len(inlineClient) == 0 {
+		return packets.ErrClientIdentifierNotValid
+	}
+
+	s.Topics.InlineSubscribe(packets.InlineSubscription{
+		Identifier: inlineClient,
```
mochi-co, line 678:
```suggestion
		ClientID: clientID,
```

server.go
```diff
@@ -661,23 +661,43 @@ func (s *Server) Publish(topic string, payload []byte, retain bool, qos byte) er
...
-// matching the given topic filter. When a message is received on the specified topic,
-// the provided handler function will be called to process the message. returning
-// true if the subscription was new.
-func (s *Server) Subscribe(topic string, handler func(topic string, pk packets.Packet)) bool {
-	return s.Topics.InlineSubscribe(packets.InlineSubscription{
-		Filter:  topic,
-		Handler: handler,
+// matching the given topic filter for a specific client. When a message is received on the specified topic,
+// the provided handler function will be called to process the message. It returns true if the subscription was new.
+func (s *Server) Subscribe(inlineClient, filter string, handler func(client string, pk packets.Packet)) error {
```
mochi-co, line 668: Because it's only an ID and not an mqtt.Client, I would recommend using `clientID` instead of `inlineClient`.

```suggestion
func (s *Server) Subscribe(clientID, filter string, handler func(client string, pk packets.Packet)) error {
```

server.go
```diff
@@ -661,23 +661,43 @@ func (s *Server) Publish(topic string, payload []byte, retain bool, qos byte) er
...
+
+	s.Topics.InlineSubscribe(packets.InlineSubscription{
+		Identifier: inlineClient,
+		Filter:     filter,
+		Handler:    handler,
 	})
+
+	// Handling retained messages.
+	for _, pkv := range s.Topics.Messages(filter) { // [MQTT-3.8.4-4]
+		handler(inlineClient, pkv)
```
mochi-co, line 685:
```suggestion
		handler(clientID, pkv)
```

server.go
```diff
@@ -661,23 +661,43 @@ func (s *Server) Publish(topic string, payload []byte, retain bool, qos byte) er
...
 
-// Unsubscribe removes the inline subscription for the specified topic.
-// This function allows you to unsubscribe from the internal subscription associated with the given topic.
-// returning true if the subscription existed.
-func (s *Server) Unsubscribe(topic string) bool {
-	return s.Topics.InlineUnsubscribe(topic)
+// Unsubscribe removes the inline subscription for the specified client and topic filter.
+// This function allows you to unsubscribe a specific client from the internal subscription
+// associated with the given topic filter. It returns true if the subscription existed.
+func (s *Server) Unsubscribe(inlineClient, filter string) error {
```
mochi-co, line 693: As above

```suggestion
func (s *Server) Unsubscribe(clientID, filter string) error {
```

server.go
```diff
@@ -661,23 +661,43 @@ func (s *Server) Publish(topic string, payload []byte, retain bool, qos byte) er
...
+	if !IsValidFilter(filter, false) {
+		return packets.ErrTopicFilterInvalid
+	} else if len(inlineClient) == 0 {
+		return packets.ErrClientIdentifierNotValid
+	}
+	s.Topics.InlineUnsubscribe(inlineClient, filter)
```
mochi-co, line 699:
```suggestion
	s.Topics.InlineUnsubscribe(clientID, filter)
```

server_test.go
```diff
@@ -3189,77 +3166,182 @@ func TestServerSubscribe(t *testing.T) {
...
+		handler func(client string, pk packets.Packet)
+		expert  error
 	}{
 		{
 			desc:    "subscribe",
 			filter:  "a/b/c",
+			client:  "client-1",
 			handler: handler,
-			wasNew:  true,
+			expert:  nil,
```
mochi-co, line 3178:
```suggestion
			expect:  nil,
```

server_test.go
```diff
@@ -3189,77 +3166,182 @@ func TestServerSubscribe(t *testing.T) {
 	tt := []struct {
 		desc    string
 		filter  string
-		handler func(topic string, pk packets.Packet)
-		wasNew  bool
+		client  string
+		handler func(client string, pk packets.Packet)
+		expert  error
```
mochi-co, line 3171: Small typo

```suggestion
		expect  error
```

server_test.go
```diff
@@ -3189,77 +3166,182 @@ func TestServerSubscribe(t *testing.T) {
...
-			wasNew:  true,
+			expert:  nil,
 		},
 		{
 			desc:    "subscribe existed",
 			filter:  "a/b/c",
+			client:  "client-1",
 			handler: handler,
-			wasNew:  false,
+			expert:  nil,
```
mochi-co, line 3185:
```suggestion
			expect:  nil,
```

server_test.go
```diff
@@ -3189,77 +3166,182 @@ func TestServerSubscribe(t *testing.T) {
...
+		},
+		{
+			desc:    "subscribe different client id",
+			filter:  "a/b/c",
+			client:  "client-2",
+			handler: handler,
+			expert:  nil,
```
mochi-co, line 3192:
```suggestion
			expect:  nil,
```

server_test.go
```diff
@@ -3189,77 +3166,182 @@ func TestServerSubscribe(t *testing.T) {
...
 		},
 		{
 			desc:    "subscribe case sensitive didnt exist",
 			filter:  "A/B/c",
+			client:  "client-1",
 			handler: handler,
-			wasNew:  true,
+			expert:  nil,
```
mochi-co, line 3199:
```suggestion
			expect:  nil,
```

server_test.go
```diff
@@ -3189,77 +3166,182 @@ func TestServerSubscribe(t *testing.T) {
...
 		},
 		{
 			desc:    "wildcard+ sub",
 			filter:  "d/+",
+			client:  "client-1",
 			handler: handler,
-			wasNew:  true,
+			expert:  nil,
```
mochi-co, line 3206:
```suggestion
			expect:  nil,
```

server_test.go
```diff
@@ -3189,77 +3166,182 @@ func TestServerSubscribe(t *testing.T) {
...
 		},
 		{
 			desc:    "wildcard# sub",
 			filter:  "d/e/#",
+			client:  "client-1",
+			handler: handler,
+			expert:  nil,
```
mochi-co, line 3213:
```suggestion
			expect:  nil,
```

server_test.go
```diff
@@ -3189,77 +3166,182 @@ func TestServerSubscribe(t *testing.T) {
...
+		},
+		{
+			desc:    "subscribe $SYS/info",
+			filter:  "$SYS/info",
+			client:  "client-1",
+			handler: handler,
+			expert:  nil,
```
mochi-co, line 3220:
```suggestion
			expect:  nil,
```

server_test.go
```diff
@@ -3189,77 +3166,182 @@ func TestServerSubscribe(t *testing.T) {
...
+		},
+		{
+			desc:    "subscribe invalied ###",
+			filter:  "###",
+			client:  "client-1",
+			handler: handler,
+			expert:  packets.ErrTopicFilterInvalid,
```
mochi-co, line 3227:
```suggestion
			expect:  packets.ErrTopicFilterInvalid,
```

server_test.go
```diff
@@ -3189,77 +3166,182 @@ func TestServerSubscribe(t *testing.T) {
...
+		},
+		{
+			desc:    "subscribe invalied " + SharePrefix + "/#",
+			filter:  SharePrefix + "/#",
+			client:  "client-1",
 			handler: handler,
-			wasNew:  true,
+			expert:  packets.ErrTopicFilterInvalid,
```
mochi-co, line 3234:
```suggestion
			expect:  packets.ErrTopicFilterInvalid,
```

server_test.go
```diff
@@ -3189,77 +3166,182 @@ func TestServerSubscribe(t *testing.T) {
...
+		},
+		{
+			desc:    "invalid handler",
+			filter:  "a/b/c",
+			client:  "client-1",
+			handler: nil,
+			expert:  packets.ErrInlineSubscriptionHandlerInvalid,
```
mochi-co, line 3241:
```suggestion
			expect:  packets.ErrInlineSubscriptionHandlerInvalid,
```

server_test.go
```diff
@@ -3189,77 +3166,182 @@ func TestServerSubscribe(t *testing.T) {
...
+		},
+		{
+			desc:    "invalid client id",
+			filter:  "a/b/c",
+			client:  "",
+			handler: handler,
+			expert:  packets.ErrClientIdentifierNotValid,
```
mochi-co, line 3248:
```suggestion
			expect:  packets.ErrClientIdentifierNotValid,
```

server_test.go
```diff
@@ -3189,77 +3166,182 @@ func TestServerSubscribe(t *testing.T) {
...
 		},
 	}
 
 	for _, tx := range tt {
 		t.Run(tx.desc, func(t *testing.T) {
-			require.Equal(t, tx.wasNew, s.Subscribe(tx.filter, tx.handler))
+			require.Equal(t, tx.expert, s.Subscribe(tx.client, tx.filter, tx.handler))
```
mochi-co, line 3254:
```suggestion
			require.Equal(t, tx.expect, s.Subscribe(tx.client, tx.filter, tx.handler))
```

server_test.go
```diff
@@ -3189,77 +3166,182 @@ func TestServerSubscribe(t *testing.T) {
...
 		})
 	}
 }
 
 func TestServerUnsubscribe(t *testing.T) {
-	handler := func(topic string, pk packets.Packet) {
+	handler := func(client string, pk packets.Packet) {
```
mochi-co, line 3260:
```suggestion
	handler := func(clientID string, pk packets.Packet) {
```

server_test.go
```diff
@@ -3189,77 +3166,182 @@ func TestServerSubscribe(t *testing.T) {
...
 		// handler logic
 	}
 
+	client := "client-1"
```
mochi-co, line 3264: Probably should be `clientID`

```suggestion
	clientID := "client-1"
```

server_test.go
```diff
@@ -3189,77 +3166,182 @@ func TestServerSubscribe(t *testing.T) {
...
 	s := New(nil)
-	exists := s.Subscribe("a/b/c/d", handler)
-	require.True(t, exists)
+	err := s.Subscribe(client, "a/b/c/d", handler)
```
mochi-co, line 3266:
```suggestion
	err := s.Subscribe(clientID, "a/b/c/d", handler)
```

server_test.go
```diff
@@ -3189,77 +3166,182 @@ func TestServerSubscribe(t *testing.T) {
...
+	require.Nil(t, err)
 
-	exists = s.Subscribe("a/b/+/d", handler)
-	require.True(t, exists)
+	err = s.Subscribe(client, "a/b/+/d", handler)
```
mochi-co, line 3269:
```suggestion
	err = s.Subscribe(clientID, "a/b/+/d", handler)
```

server_test.go
```diff
@@ -3189,77 +3166,182 @@ func TestServerSubscribe(t *testing.T) {
...
+	require.Nil(t, err)
 
-	exists = s.Subscribe("d/e/f", handler)
-	require.True(t, exists)
+	s.Subscribe(client, "#", handler)
```
mochi-co, line 3272:
```suggestion
	s.Subscribe(clientID, "#", handler)
```

server_test.go
```diff
@@ -3189,77 +3166,182 @@ func TestServerSubscribe(t *testing.T) {
...
+	require.Nil(t, err)
 
-	s.Subscribe("d/e/f", handler)
-	require.True(t, exists)
+	err = s.Unsubscribe(client, "a/b/c/d")
```
mochi-co, line 3275:
```suggestion
	err = s.Unsubscribe(clientID, "a/b/c/d")
```

server_test.go
```diff
@@ -3189,77 +3166,182 @@ func TestServerSubscribe(t *testing.T) {
...
+	require.Nil(t, err)
 
-	s.Subscribe("#", handler)
-	require.True(t, exists)
+	err = s.Unsubscribe(client, "fdasfdas/dfsfads/sa")
```
mochi-co, line 3278:
```suggestion
	err = s.Unsubscribe(clientID, "fdasfdas/dfsfads/sa")
```

server_test.go
```diff
@@ -3189,77 +3166,182 @@ func TestServerSubscribe(t *testing.T) {
...
+	require.Nil(t, err)
 
-	ok := s.Unsubscribe("a/b/c/d")
-	require.True(t, ok)
+	err = s.Unsubscribe(client, "#/#")
```
mochi-co, line 3281:
```suggestion
	err = s.Unsubscribe(clientID, "#/#")
```

topics.go
```diff
@@ -186,58 +186,58 @@ func (s *SharedSubscriptions) GetAll() map[string]map[string]packets.Subscriptio
 	return m
 }
 
-// InlineSubscriptions represents a map of internal subscriptions keyed on the topic filter.
+// InlineSubscriptions represents a map of internal subscriptions keyed on client.
 type InlineSubscriptions struct {
-	subs map[string]packets.InlineSubscription
+	internal map[string]packets.InlineSubscription
```
mochi-co, line 191: Good change 👍🏻

topics.go
```diff
@@ -361,15 +361,15 @@ func (x *TopicsIndex) InlineSubscribe(subscription packets.InlineSubscription) b
 
 	var existed bool
 	n := x.set(subscription.Filter, 0)
-	_, existed = n.inlineSubscriptions.Get(subscription.Filter)
+	_, existed = n.inlineSubscriptions.Get(subscription.Identifier)
```
mochi-co, line 364:
```suggestion
	_, existed = n.inlineSubscriptions.Get(subscription.ClientID)
```

mochi-co:
@werbenhu I got started on a review but it got a bit much in terms of little changes, I clicked `cancel review` but it seems to have attached the issues. Please disregard them and instead see my notes to follow

mochi-co:
@werbenhu 
This was always going to be a technically tricky feature to implement, so please excuse my feedback as I'm still trying to determine the best way for this to be implemented. 

Easy stuff:
- Small typo in `TestServerSubscribe(t *testing.T) {` - references to `expert` should be `expect`
- The move from `subs` to `internal` in `InlineSubscriptions` is good, and is consistent with the rest of the code.
- `TestServerSubscribe` should also test that a new subscription from one ID shouldn't overwrite existing subscriptions from other IDs.
- `TestServerUnsubscribe` should also test that unsubscriptions from one ID shouldn't unsubscribe subscriptions from other IDs.

Harder stuff:
- The `InlineSubscription` struct currently uses the `Identifier` field as the key for the subscription, and is populated with the `client id`. However, we can't use `Identifier` as this refers to a specific MQTTv5 feature - Subscription Identifiers (int), so it becomes confusing.
- Similarly, places where we have `client` vars which use strings - these should probably be `clientID` as `client` typically refers to the client struct in all other places in the code (which is usually just `cl`).
- If this should trigger `hook.OnSubscribe` - I am inclined to say yes, if only because it would also resolve the above issues - `cl *Client` could be passed through instead of `clientID`. While it's a bit more work for the developer to make an `NewClient` and pass that in, I think doing so keeps us open for non-breaking extendability in the future, and makes the internal server code more consistent.

Having said that, once we do this, then we are 95% of the way back to the initial implementation of using Inline Clients, except that we are storing the subscriptions in a new map instead of with the others. As for me, I think this PR has a lot of good work in it that gets us most of the way there. Maybe it could be simpler if we combine the two.

This might be a bit controversial, but here's what I would do:

* Omit all the inline subscription maps. If we are going to use the subscriptions in exactly the same way as normal subscriptions, I'm not sure it makes sense to use them when 99% of the functionality will be identical, even if I do appreciate the simplicity.
* Add a `Handler` to `packets.Subscription`
```go
// Subscription contains details about a client subscription to a topic filter.
type Subscription struct {
	ShareName         []string
	Filter            string
	Identifier        int
	Identifiers       map[string]int
	RetainHandling    byte
	Qos               byte
	RetainAsPublished bool
	NoLocal           bool
	FwdRetainedFlag   bool // true if the subscription forms part of a publish response to a client subscription and packet is retained.
	Handler           func(cl any, sub Subscription, packet Packet) // cl to be asserted as mqtt.Client (avoid import cycle)
}
```

* Pass NewClient() inline client to Subscribe and Unsubscribe. Using our modified `Subscription` with `Handler`, `processSubscribe`. Index the subscription as normal. 
* In `(s *Server) publishToClient`, after topic aliases but before Qos>0:
```go
if sub.Handler != nil {
    go sub.Handler(cl, sub, out)
    return out, nil
}

if out.FixedHeader.Qos > 0 {
```

Effectively I am talking about using the handler concepts and tests from @werbenhu's PR and putting them directly into packet.Subscription as a property of a Subscription, rather than maintaining it all as a separate map. I believe this would be the highest performance and lowest technical overhead option. I'd be interested to see what you think of this route - @dgduncan @thedevop @werbenhu.

werbenhu:
I've come up with a few ideas that I'd like to put forward for everyone's consideration and gather some opinions @dgduncan @thedevop @werbenhu.

 - Perhaps we don't need to support subscriptions for multiple clients, similar to server.Publish(), by using a client with the clientid set to "inline" as the fixed client for all inline subscriptions.
We can use the Identifier of the Subscription as the key to manage all inline subscriptions. For example, it could be implemented like this:

```golang
const (
	Version                       = "2.3.0" // the current server version.
	defaultSysTopicInterval int64 = 1       // the interval between $SYS topic publishes
	LocalListener                 = "local"
	InlineClientId                = "inline"
)

//...

func (s *Server) Subscribe(filter string, id int, handler InlineSubFn) error {

	if handler == nil {
		return packets.ErrInlineSubscriptionHandlerInvalid
	}

	reasonCodes := make([]byte, 1)
	if !IsValidFilter(filter, false) {
		reasonCodes[0] = packets.ErrTopicFilterInvalid.Code
	}

	cl := s.NewClient(nil, LocalListener, InlineClientId, true)
	if _, ok := s.Clients.Get(cl.ID); !ok {
		// add the inline client to the server if it doesn't exist, otherwise, it won't be eligible for topic matching.
		s.Clients.Add(cl)
	}

	pk := s.hooks.OnSubscribe(cl, packets.Packet{ // subscribe like a normal client.
		Origin:      cl.ID,
		FixedHeader: packets.FixedHeader{Type: packets.Subscribe},
		Filters: packets.Subscriptions{
			{Filter: filter, Identifier: id},
		},
	})

	sub := InlineSubscription{
		Subscription: packets.Subscription{
			Identifier: id,
			Filter:     filter,
		},
		Handler: handler,
	}
	s.Topics.InlineSubscribe(sub)

	s.hooks.OnSubscribed(cl, pk, reasonCodes)
	// Handling retained messages.
	for _, pkv := range s.Topics.Messages(filter) { // [MQTT-3.8.4-4]
		handler(cl, sub.Subscription, pkv)
	}
	return nil
}

func (s *Server) Unsubscribe(filter string, id int) error {
	if !IsValidFilter(filter, false) {
		return packets.ErrTopicFilterInvalid
	}

	if cl, ok := s.Clients.Get(InlineClientId); ok {
		pk := s.hooks.OnUnsubscribe(cl, packets.Packet{ // unsubscribe like a normal client.
			Origin:      cl.ID,
			FixedHeader: packets.FixedHeader{Type: packets.Unsubscribe},
			Filters: packets.Subscriptions{
				{Filter: filter},
			},
		})
		s.Topics.InlineUnsubscribe(id, filter)
		s.hooks.OnUnsubscribed(cl, pk)
	}

	return nil
}
```

 - Furthermore, I believe we shouldn't tightly couple inline subscriptions with packets.
I suggest creating an InlineSubscription struct that inherits from packets.Subscription, This way, we can separate Handler from packets, hook, and maintain the characteristics of Subscription to keep us open for non-breaking extendability in the future.

```golang
// topics.go

// ClientSubscriptions is a map of aggregated subscriptions for a client.
type ClientSubscriptions map[string]packets.Subscription

type InlineSubscription struct {
	packets.Subscription
	Handler InlineSubFn
}
```

werbenhu:
Going a step further, we can directly store an inlineClient within the Server struct, creating this client when the server starts, like this:
```
type Server struct {
	Options      *Options             // configurable server options
	Listeners    *listeners.Listeners // listeners are network interfaces which listen for new connections
	Clients      *Clients             // clients known to the broker
	Topics       *TopicsIndex         // an index of topic filter subscriptions and retained messages
	Info         *system.Info         // values about the server commonly known as $SYS topics
	loop         *loop                // loop contains tickers for the system event loop
	done         chan bool            // indicate that the server is ending
	Log          *zerolog.Logger      // minimal no-alloc logger
	hooks        *Hooks               // hooks contains hooks for extra functionality such as auth and persistent storage.
	InlineClient *Client			  // InlineClient is a special client used for inline subscriptions and inline Publish.
}
```

werbenhu:
@mochi-co I've made some changes and temporarily removed the test cases to make it easier for everyone to review the proposed modifications. Please disregard the build failures caused by the absence of test cases for now.

dgduncan:
@werbenhu @mochi-co You two are far more knowledgable about the inline client and most likely the processes of the server itself so for the very technical aspects I default to you; however, my opinion for a few points above.

If I am reading this right, I absolutely agree that there should be a single inline client responsible for publishing and subscribing. It makes the most sense to me to think of that client as a very special admin client where only one can exist

Also I will say that I do like the idea of having the inline client as part of the server struct itself. Something about it makes it seem very "clean" and with it you may be easier to lock all of the inline client stuff behind a `capability` but perhaps I am stretching there.

server.go
```diff
@@ -661,6 +665,65 @@ func (s *Server) Publish(topic string, payload []byte, retain bool, qos byte) er
...
+		Origin:      s.inlineClient.ID,
+		FixedHeader: packets.FixedHeader{Type: packets.Subscribe},
+		Filters:     packets.Subscriptions{subscription},
+	})
+
+	inlineSubscription := InlineSubscription{
+		Subscription: subscription,
+		Handler:      handler,
+	}
+	s.Topics.InlineSubscribe(inlineSubscription)
```
mochi-co, line 693: Not sure we should actually subscribe if `reasonCodes[0] == packets.ErrTopicFilterInvalid.Code` ?

topics.go
```diff
@@ -633,23 +746,25 @@ func IsValidFilter(filter string, forPublish bool) bool {
...
-	subscriptions *Subscriptions       // a map of subscriptions made by clients to this ending address
-	shared        *SharedSubscriptions // a map of shared subscriptions keyed on group name
-	retainPath    string               // path of a retained message
-	sync.Mutex                         // mutex for when making changes to the particle
+	key                 string               // the key of the particle
+	parent              *particle            // a pointer to the parent of the particle
+	particles           particles            // a map of child particles
+	subscriptions       *Subscriptions       // a map of subscriptions made by clients to this ending address
+	shared              *SharedSubscriptions // a map of shared subscriptions keyed on group name
+	inlineSubscriptions *InlineSubscriptions
```
mochi-co, line 754: Needs a comment, if only to make everything align correctly 😅

mochi-co (approved):
While bold, I think this is the most correct solution so far. I really enjoy how clean it is, and it tidies up the whole concept of inline clients into something a bit more definitive and less hacky. I agree with you and @dgduncan that the inline client is more appropriate as a single medium for data input/output, and this removes any ambiguity to its purpose. 

Additionally, those who need more sophisticated client features can potentially use the paho client libraries directly and connect locally.

For me, this is a really great step in the right direction. I think you solved it @werbenhu!

I have approved the PR, pending tests and finalisation.

werbenhu:
@mochi-co Thanks! I've gained a lot of inspiration from your suggestions and the code you submitted in PR#283. I will be adding some test cases later on. For me, writing test cases is often more challenging than writing the implementation code.

werbenhu:
@mochi-co @dgduncan I have just completed the modifications and added comprehensive test cases.

dgduncan (approved):
Approved!

thedevop:
Now that we're moving to a permanent inline client (which is great), we would also need to review the Publish path so it fully account for the inline client. For example, currently if Publish with QOS 2, the quota will be reduced but will not have a chance to increase back (not an issue when a new inline client is created). Perhaps we can change the following in processPublish:
```go
	if pk.FixedHeader.Qos == 0 {
		s.publishToSubscribers(pk)
		s.hooks.OnPublished(cl, pk)
		return nil
	}
```
to 
```go
	if pk.FixedHeader.Qos == 0 || cl.Net.Inline {
		s.publishToSubscribers(pk)
		s.hooks.OnPublished(cl, pk)
		return nil
	}
```

werbenhu:
@thedevop I think you've raised a very good question, which is how to handle PUBREC and PUBREL when inlinceClient acts as a publisher. Your code would set the QoS of all inline PUBLISH packets to 0, making the 'qos' parameter in 'server.Publish()' meaningless.

werbenhu:
@thedevop Maybe you're right. When inlineClient acts as a publisher, the server treats it as qos=0, but for the client, the qos of this PUBCOMP message is still 2.

werbenhu:
@thedevop If you don't mind, I'll include this change in this PR. :heart:

mochi-co:
I will try to give this a final review later this evening - note we have a small conflict in `server.go` because of slog merge.

thedevop:
> @thedevop You 're right. When inlineClient acts as a publisher, the server treats it as qos=0 and doesn't need to handle PUBREC and PUBREL. However, for the client, the qos of this PUBLISH message is still 2. Nice catch!
> 
> If you don't mind, I'll include this change in this PR. ❤️

Yes, please.

werbenhu:
@mochi-co Sorry for the delayed submission. I had something to attend to last night.

dgduncan:
@werbenhu Just wanted to check to see if you believe this is ready to review before I go ahead and take a look :)

werbenhu:
@dgduncan Yes, that's the final.

werbenhu:
@dgduncan Nothing to submit. It's ready for review. Please go ahead. Thanks!

dgduncan (approved):
Approved with a question. I know I mentioned it in the PR discussion and I could have sworn that @thedevop mentioned this but I can not find it. Do we think there is a any strong reasoning to lock the inline client behind a `capability` as to allow it to skip a few steps when running? Or does that seem like a bit too much effort/complexity without a lot of payoff?

thedevop:
@dgduncan, I think you were referring to:
3. Has the least impact on the server performance when no inline subscriptions are used.

Based. on the current implementation, the impact should be relatively small. So probably does not warrant locking it behind some flag.

mochi-co (approved):

mochi-co:
I've given it a proper line by line review. There's one thing I'm not 100% on, but it's not serious enough to prevent us from merging this feature:

- `server.Clients also includes a server.inlineClient, so the total number of clients here should be increased by 1`
For me, this is a compelling reason to lock inline clients behind a capability. I'm not super comfortable with the broker always advertising 1 connected client by default, I feel this could be confusing for users who aren't aware of the inline client feature.

I may have a little look at locking this behind a capability before release. 

Otherwise, congratulations @werbenhu and many thanks for all your hard work! I really love the direction we are going in!

mochi-co:
Released in [v2.4.0](https://github.com/mochi-mqtt/server/releases/tag/v2.4.0) - Really excellent work on this @werbenhu, it's a brilliant improvement to the project and I'm very happy with the new direction for inline clients. Feels much 'cleaner' and tidier. Definitely a contribution to be proud of!

Issue #286, opened 2023-09-01
acl rules do not take effect when published to the client


xyzj:
This is my acl rules:
Ledger: &auth.Ledger{
	Auth: auth.AuthRules{
		{Username: "user1", Password: "pwd1", Allow: true},
		{Username: "user2", Password: "pwd2", Allow: true},
		{Username: "user3", Password: "pwd3", Allow: true},
	},
	ACL: auth.ACLRules{
		{
			Username: "user1",
			Filters: auth.Filters{
				"wo/#": auth.WriteOnly,
			},
		},
		{
			Username: "user2",
			Filters: auth.Filters{
				"wo/#": auth.WriteOnly,
				"ro/#": auth.WriteOnly,
			},
		},
	},
}
use1 can publish message with "wo/#" but can not receive message with "wo/#", When user 1's subscription is set to "wo/#", it works fine. but ,when use1's subscription is set to "#", it can receive the message with "wo/#" from user2.
I think aclcheck should also be added in publishToClient

mochi-co:
Fixed and released in [v2.4.0](https://github.com/mochi-mqtt/server/releases/tag/v2.4.0) - Thank you @xyzj - a wonderful first contribution!

PR #287, opened 2023-09-01
add aclcheck in publishToClient

xyzj:
As issue #286, I do some change in server.publishToClient

### DIFF
--- a/server.go
+++ b/server.go
@@ -26,8 +26,8 @@ import (
 )
 
 const (
-	Version                       = "2.3.0"  // the current server version.
-	defaultSysTopicInterval int64 = 1        // the interval between $SYS topic publishes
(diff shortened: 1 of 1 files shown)


mochi-co (approved):
This makes sense to me, and it's a good catch! We can probably move the `out := ` declaration below the ACL check but it's a minor issue:
```go
if !s.hooks.OnACLCheck(cl, pk.TopicName, false) {
    return pk, packets.ErrNotAuthorized
}
out := pk.Copy(false)
```

Additionally, a new test case is required (`TestPublishToClientServerTopicAlias` could be used as a starting point). However, happy to add this later before release.

dgduncan (approved):
This looks good to me!

Issue #290, opened 2023-09-06
When a MQTTv311 client tries to publish with qos2, and fails OnACLCheck, the client hangs forever and the connection stays open. Should we disconnect the client instead?


SimonMacIntyre:
http://docs.oasis-open.org/mqtt/mqtt/v3.1.1/os/mqtt-v3.1.1-os.html#_Toc398718042

> If a Server implementation does not authorize a PUBLISH to be performed by a Client; it has no way of informing that Client. It MUST either make a positive acknowledgement, according to the normal QoS rules, or close the Network Connection [MQTT-3.3.5-2].

I encountered this when testing my implementation and testing a `publish qos2` which `succesfully Connects (authenticates), but fails OnACLCheck`, in my case using Python Paho client 3.1.1, the publish fails but the client hangs forever on the publish. 

When I looked into the broker code it appears on failure of ACL check that it returns `nil` with no handling. 

Please let me know if I am missing anything or misinformed!

SimonMacIntyre:
This case is the same using client of MQTTv5 too

mochi-co:
I believe you are correct @SimonMacIntyre, this is a bug, at least for MQTTv3. If I remember, for v5 there is no provision to disconnect on ACL errors as some servers may wish to obscure their restrictions. I may be wrong though.

SimonMacIntyre:
I'll take a quick peek! I see these in the v5 spec:
> The Client or Server sending the PUBACK packet MUST use one of the PUBACK Reason Codes [MQTT-3.4.2-1]
>The Client or Server sending the PUBREC packet MUST use one of the PUBREC Reason Code values. [MQTT-3.5.2-1]

Same  PUBACK reason code for both would be:
>135 | 0x87 | Not authorized | The PUBLISH is not authorized.

For qos0, it appears to be handled correctly!

dgduncan:
So from reading that it seems like the spec for v3.1.1 should be to either ack and not publish or disconnect. A disconnect seems like the preferred solution as it makes it obvious there is a breach in the server rules; however, I can absolutely imagine a world in which a user of the server may not have control over some functionality of their MQTT client because it may be third part and a disconnection may make the server useless to them. Perhaps a capability could be in order to adjust functionality but that may just add unwanted complexity

SimonMacIntyre:
Personally I lean towards the disconnect, over the ack and not publish. 

[...]

PR #292, opened 2023-09-07
Disconnect or return ack if unauthorized publish

thedevop:
Addresses #290.


dgduncan (approved):
Approved.

server_test.go
```diff
@@ -1421,14 +1442,65 @@ func TestServerProcessPublishInvalidTopic(t *testing.T) {
...
+		{"v5_QOS0", 5, *packets.TPacketData[packets.Publish].Get(packets.TPublishBasicMqtt5).Packet,
+			nil, nil, false},
+		{"v5_QOS1", 5, *packets.TPacketData[packets.Publish].Get(packets.TPublishQos1Mqtt5).Packet,
+			nil, packets.TPacketData[packets.Puback].Get(packets.TPubrecMqtt5NotAuthorized).RawBytes, false},
+		{"v5_QOS2", 5, *packets.TPacketData[packets.Publish].Get(packets.TPublishQos2Mqtt5).Packet,
+			nil, packets.TPacketData[packets.Pubrec].Get(packets.TPubrecMqtt5NotAuthorized).RawBytes, false},
+	}
+
+	for _, tt := range tests {
+		tt := tt
```
mochi-co, line 1469: @thedevop curious to the reason for this line

mochi-co (approved):
Quality PR, runs well, looks good to me. Approved!

mochi-co:
Thank you very much @thedevop - your expertise is greatly appreciated as always! 💯

server_test.go
thedevop: This is to address this issue: https://github.com/golang/go/wiki/CommonMistakes.

Although as it is, tt := tt is not needed because these tests are not running in parallel, but out of habit I add that as often I do run test cases in parallel (only one line change).

mochi-co: Very interesting, thank you!

PR #296, opened 2023-09-09
Add MinimumKeepalive

thedevop:


werbenhu:
@thedevop You cannot directly modify the server's keepalive because you cannot synchronize it with the client. The client also needs to use this parameter to send ping packets.

thedevop:
For MQTT v5, the server will notify the client to use the MinimumKeepalive. But unfortunately for prior versions, there is no mechanism to notify the client. So we can either add a check for MQTT v5 only, or intentional not honoring the client setting.

thedevop:
@werbenhu , the more I think about it, this maybe unnecessary, especially this can not be enforced in < MQTT v5 without breaking the spec. Should I just close this PR?

werbenhu:
@thedevop I agree with @dgduncan  suggestion to add a Warn-level log, which would be better.

Issue #227, opened 2023-05-15
OpenTelemetry Traces & Logging


IshanDaga:
## Feature Description
OTEL Plugin / Hook-based system to track a packet's lifecycle through the broker.
Would like observability into how my hooks are dealing with connecting clients and the lifecycle of the client / packet being processed.

## Current Thinking
- Each packet could be it's own Trace, and every Hook interacting with said packet / transaction would have it's own span within that trace.
- Potentially use the `packetId` as the Trace identifier

IshanDaga:
Seeing this mentioned elsewhere as well, have had some time to think about this - in my current environment, it is important for me to trace the packet outside of the "context" of the broker as well ( client -> broker -> queue -> service ) till the packet is processed by a consumer / subscriber.

That being said, what we're doing internally is inserting a `trace_id` to the packet using the existing `OnPacketRead` hook.
This `trace_id` is generated using the `client_id` and the current epoch time (ms).

Each hook then adds a span ( or span events / logs ) to this trace. This trace (for now) is closed by the service, basically the last entity that needs to deal with the packet payload.

Thoughts / suggestions @mochi-co ?

IshanDaga:
Update here :

I have quite successfully been using OnPacketRead and OnPacketProcessed Hooks to add / remove TraceIds from packet payloads

I have one request, could we also expose the entire packet to the OnACLCheck Hook ? 

Examples of how I'm doing things below :
#### Helper Functions
```
package utils

import (
	"bytes"

	"go.opentelemetry.io/otel/trace"
)

var TraceSeparator = []byte(":TID:")

// Attaches a trace id to a payload
func AttachTraceId(payload []byte, traceId []byte) []byte {
	tracePayload := append(TraceSeparator, traceId...)
	return append(payload, tracePayload...)
}

// Parse a trace id from a raw byte array
func ParseTraceIdFromBytes(traceId []byte) trace.TraceID {
	tId, err := trace.TraceIDFromHex(string(traceId))
	if err != nil {
		return trace.TraceID{}
	}
	return tId
}

// Removes the trace id from a payload (if it exists)
func RemoveTraceIdFromPayload(payload []byte) []byte {
	index := bytes.Index(payload, TraceSeparator)
	if index != -1 {
		return payload[:index]
	}
	return payload
}

func TraceIdFromPacket(payload []byte) trace.TraceID {
	index := bytes.Index(payload, TraceSeparator)
	if index != -1 {
[...]

PR #297, opened 2023-09-09
[WIP] OTEL Observability

IshanDaga:
### Feature
Using OTEL lib to instrument the server. Happy to take suggestions on things that we should track, metrics to add, and how to allow maximum configurability for the Tracer.

### Tasks
- [x] Add context object to each packet
- [x] Add OTEL config options to `server.Options` struct
- [ ] Add spans for client events
- [ ] Add metrics tracking using OTEL Meters API
- [ ] Instrument slog handler

### Pending decisions
- [ ] should we add another external dependency like https://github.com/ttys3/slogx to internally instrument slog, or should we leave that to the user to implement within their own hook implementations ?


IshanDaga:
@dgduncan @mochi-co 
Another approach to this could be to have a `StartTrace` hook that is called after `receivePacket`
We could also add in a trace-modifier hook of sorts, that allows someone to insert the TraceID into the header for MQTTv5 or modify the packet to include TraceID bytes with a separator like in #227 

Looking for initial comments and thoughts :)

IshanDaga:
@mochi-co hey I'm having trouble with the test cases involving sample packets.
Most are failing on the `require.Equal` since now a packet from the server has a context, but the sample packet doesn't.

There's quite a few of these cases, any suggestions on how I could move past this without breaking the current testing method ?

mochi-co:
@IshanDaga SImplest solution would be to only populate context if OTEL tracing is enabled:

```diff
Index: clients.go
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/clients.go b/clients.go
--- a/clients.go	(revision 9d6c78915853d616b25757776b6945885b48df22)
+++ b/clients.go	(date 1697972510469)
@@ -454,7 +454,9 @@
 func (cl *Client) ReadPacket(fh *packets.FixedHeader) (pk packets.Packet, err error) {
 	atomic.AddInt64(&cl.ops.info.PacketsReceived, 1)
 
-	pk.Ctx, pk.Cancel = context.WithCancel(cl.State.open)
+	if cl.ops.options.OTELTracing {
+		pk.Ctx, pk.Cancel = context.WithCancel(cl.State.open)
+	}
 
 	pk.ProtocolVersion = cl.Properties.ProtocolVersion // inherit client protocol version for decoding
 	pk.FixedHeader = *fh
```

IshanDaga:
@mochi-co makes sense, thanks !

mochi-co:
@IshanDaga This PR is a bit out of date, I recommend closing it and opening a new one when there's more changes 👍🏻

IshanDaga:
@mochi-co makes sense, closing for now, will get back on this as soon as i have a little more free time

Issue #298, opened 2023-09-13
badger crashed on server  Close


fangdingjun:
Hi,

I use badger to store the data, when I try to restart the server, sometimes crashed.

I make a demo that can  reproduce the issue.

test step

1. start the server

2. start some clients connect to server and keep sending messages
I use jmeter with mqtt plugin, 10 threads, infinite loop

3. press CTRL+C to close the server

after ctrl+c the server crashed and print the stack

version

    golang 1.20.4  linux/amd64
    mochi-mqtt  v2.3.0

this is the demo code

    package main
    
    import (
            "os"
            "os/signal"
            "syscall"
    
            "github.com/mochi-mqtt/server/v2"
            "github.com/mochi-mqtt/server/v2/hooks/auth"
            "github.com/mochi-mqtt/server/v2/hooks/storage/badger"
            "github.com/mochi-mqtt/server/v2/listeners"
    )
    
    func main() {
            server := mqtt.New(nil)
            server.AddListener(listeners.NewTCP("t1", ":1883", nil))
            server.AddHook(new(auth.AllowHook), nil)
            os.MkdirAll("./data", 0755)
            server.AddHook(new(badger.Hook), &badger.Options{Path: "./data"})
            server.Serve()
    
            sigCh := make(chan os.Signal, 1)
            signal.Notify(sigCh, syscall.SIGINT, syscall.SIGTERM)
    
            <-sigCh
    
            server.Close()
    }

    

this is the trace

    
      2:01PM INF stopping hook hook=allow-all-auth
      2:01PM INF stopping hook hook=badger-db
      2:01PM ERR failed to upsert client data error="Writes are blocked, possibly due to DropAll or Close" data={"clean":true,"id":"conn_0735f94a355d46d9b1","listener":"t1","properties":{"authenticationData":null,"authenticationMethod":"","maximumPacketSize":0,"receiveMaximum":0,"requestProblemInfo":0,"requestProblemInfoFlag":false,"requestResponseInfo":0,"sessionExpiryInterval":0,"sessionExpiryIntervalFlag":false,"topicAliasMaximum":0,"user":null},"protocolVersion":4,"remote":"192.168.56.1:50083","t":"CL","username":null,"will":{"flag":0,"payload":null,"qos":0,"retain":false,"topicName":"","user":null,"willDelayInterval":0}} hook=badger-db
[...]

PR #299, opened 2023-09-13
For #298

werbenhu:
For #298,  Just see if this change is effective, and if it is, I think I should know the reason.

### DIFF
--- a/clients.go
+++ b/clients.go
@@ -372,7 +372,7 @@ func (cl *Client) waitForShutdownSignal() {
 	}
 }
 
-// sendShutdownSignal synchronously notifies other goroutines that the connection has been shutdown.
+// sendShutdownSignal notifies other goroutines that the connection has been shutdown.
 func (cl *Client) sendShutdownSignal() {
 	if cl.StopCause() != packets.ErrSessionTakenOver &&
 		cl.State.shutdownSignal != nil {
--- a/clients.go
+++ b/clients.go
@@ -150,7 +150,7 @@ type ClientState struct {
 	outboundQty     int32                // number of messages currently in the outbound queue
 	Keepalive       uint16               // the number of seconds the connection can wait
 	ServerKeepalive bool                 // keepalive was set by the server
-	waitShutdown    chan any
+	waitShutdown    chan any             // channel for synchronizing shutdown
 }
 
 // newClient returns a new instance of Client. This is almost exclusively used by Server
@@ -358,12 +358,14 @@ func (cl *Client) Read(packetHandler ReadFn) error {
 	}
 }
 
+// Shutdown synchronously notifies other goroutines that the connection has been fully closed.
 func (cl *Client) Shutdown() {
 	if cl.StopCause() != packets.ErrSessionTakenOver {
 		close(cl.State.waitShutdown)
 	}
 }
 
+// StopAndWaitShutdown instructs the client to stop and waits for the shutdown to complete.
 func (cl *Client) StopAndWaitShutdown(err error) {
 	cl.Stop(err)
 	<-cl.State.waitShutdown
--- a/clients.go
+++ b/clients.go
@@ -150,7 +150,7 @@ type ClientState struct {
 	outboundQty     int32                // number of messages currently in the outbound queue
 	Keepalive       uint16               // the number of seconds the connection can wait
 	ServerKeepalive bool                 // keepalive was set by the server
-	waitShutdown    chan any             // channel for synchronizing shutdown
+	shutdownSignal  chan any             // channel for synchronizing shutdown
 }
 
 // newClient returns a new instance of Client. This is almost exclusively used by Server
@@ -166,7 +166,6 @@ func newClient(c net.Conn, o *ops) *Client {
 			cancelOpen:    cancel,
 			Keepalive:     defaultKeepalive,
 			outbound:      make(chan *packets.Packet, o.options.Capabilities.MaximumClientWritesPending),
-			waitShutdown:  make(chan any),
 		},
 		Properties: ClientProperties{
 			ProtocolVersion: defaultClientProtocolVersion, // default protocol version
@@ -358,17 +357,27 @@ func (cl *Client) Read(packetHandler ReadFn) error {
 	}
 }
 
-// Shutdown synchronously notifies other goroutines that the connection has been fully closed.
-func (cl *Client) Shutdown() {
-	if cl.StopCause() != packets.ErrSessionTakenOver {
-		close(cl.State.waitShutdown)
+// OpenShutdownSignal initializes a shutdown signal channel for the client.
+// This channel is used to wait for a shutdown notification.
+func (cl *Client) OpenShutdownSignal() {
+	cl.State.shutdownSignal = make(chan any)
+}
+
+// WaitForShutdownSignal waits for the shutdown signal to be received.
+func (cl *Client) WaitForShutdownSignal() {
+	if cl.State.shutdownSignal != nil {
+		<-cl.State.shutdownSignal
 	}
 }
 
-// StopAndWaitShutdown instructs the client to stop and waits for the shutdown to complete.
-func (cl *Client) StopAndWaitShutdown(err error) {
-	cl.Stop(err)
-	<-cl.State.waitShutdown
+// SendShutdownSignal synchronously notifies other goroutines that the connection has been shutdown.
+func (cl *Client) SendShutdownSignal() {
+	if cl.StopCause() != packets.ErrSessionTakenOver &&
+		cl.State.shutdownSignal != nil {
+
+		close(cl.State.shutdownSignal)
+		cl.State.shutdownSignal = nil
+	}
 }
 
 // Stop instructs the client to shut down all processing goroutines and disconnect.
--- a/clients.go
+++ b/clients.go
@@ -366,9 +366,16 @@ func (cl *Client) initShutdownSignal() {
 }
 
 // waitForShutdownSignal waits for the shutdown signal to be received.
-func (cl *Client) waitForShutdownSignal() {
+func (cl *Client) waitForShutdownSignal(ctx context.Context) {
 	if cl.State.shutdownSignal != nil {
-		<-cl.State.shutdownSignal
+		// Wait for the client to signal that it has completed its shutdown process.
+		select {
+		case <-cl.State.shutdownSignal:
+			// Shutdown signal received from the client.
+		case <-ctx.Done():
+			// Context canceled, exit waiting for the shutdown signal.
+			cl.ops.log.Error("Connection closure timed out. Adjusting the ShutdownClientsWorkerNum and ShutdownClientsTimeout in Capabilities to resolve it.")
+		}
 	}
 }
 
--- a/clients.go
+++ b/clients.go
@@ -357,9 +357,9 @@ func (cl *Client) Read(packetHandler ReadFn) error {
 	}
 }
 
-// openShutdownSignal initializes a shutdown signal channel for the client.
+// initShutdownSignal initializes a shutdown signal channel for the client.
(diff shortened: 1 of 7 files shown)


werbenhu:
@thedevop @mochi-co @dgduncan  as we can see in #298, the server.Close() method actually closes all connections asynchronously. This leads to hook.OnDisconnect() and hook.Stop() being executed in separate goroutines, causing concurrent conflicts in the hook's database operations.

I believe we should support a synchronous way of disconnecting. In this mode, after disconnecting a client, we should wait for hook.OnDisconnect() to complete.

In the server.closeListenerClients() function, we should use a synchronous approach to wait for all connections to be completely shut down.
```golang
// closeListenerClients closes all clients on the specified listener.
func (s *Server) closeListenerClients(listener string) {
	clients := s.Clients.GetByListener(listener)
	var wg sync.WaitGroup
	for _, cl := range clients {
		wg.Add(1)
		go func(c *Client) {
			s.DisconnectClient(c, packets.ErrServerShuttingDown, true)
			wg.Done()
		}(cl)
	}
	wg.Wait()
}```

I've submitted a preliminary version of the code in #299 to see if anyone has any better suggestions or feedback.

**Comment by dgduncan** [comment]
@werbenhu This looks great! QQ is there any risk that due to some client or network problem `s.DisconnectClient(c, packets.ErrServerShuttingDown, true)` may hold up the graceful shutdown and prevent the server from shutting down all together?

**Comment by werbenhu** [comment]
From my current testing, everything seems fine. However, perhaps I haven't considered all aspects thoroughly, so I'd appreciate input from you @dgduncan @thedevop @mochi-co .

**Comment by thedevop** [comment]
I apologize this past week and the next few weeks I'll be busy on some other tasks.

Currently:
1. For each new connection, a Go routine is used for EstablishConnection
2. In EstablishConnection, a NewClient uses another Go routine is used for WriteLoop

File: clients.go [hunk 1]
```diff
@@ -356,6 +357,29 @@ func (cl *Client) Read(packetHandler ReadFn) error {
 	}
 }
 
+// openShutdownSignal initializes a shutdown signal channel for the client.
+// This channel is used to wait for a shutdown notification.
+func (cl *Client) openShutdownSignal() {
+	cl.Lock()
+	defer cl.Unlock()
+	cl.State.shutdownSignal = make(chan any)
+}
```
thedevop, line 366: Initialize shutdownSignal as part of NewClient?
Change name to shutdownComplete?

clients.go
```diff
@@ -356,6 +357,29 @@ func (cl *Client) Read(packetHandler ReadFn) error {
...
+
+// waitForShutdownSignal waits for the shutdown signal to be received.
+func (cl *Client) waitForShutdownSignal() {
+	if cl.State.shutdownSignal != nil {
+		<-cl.State.shutdownSignal
+	}
+}
```
thedevop, line 373: If cl.State.shutdownSignal is initialized in NewClient, we can remove if check.

clients.go
```diff
@@ -356,6 +357,29 @@ func (cl *Client) Read(packetHandler ReadFn) error {
...
+
+// sendShutdownSignal notifies other goroutines that the connection has been shutdown.
+func (cl *Client) sendShutdownSignal() {
+	if cl.StopCause() != packets.ErrSessionTakenOver &&
+		cl.State.shutdownSignal != nil {
+		close(cl.State.shutdownSignal)
+	}
+}
```
thedevop, line 381: Always close cl.State.shutdownSignal?

server.go
```diff
@@ -383,6 +384,8 @@ func (s *Server) attachClient(cl *Client, listener string) error {
 	}
 
 	s.hooks.OnSessionEstablished(cl, pk)
+	cl.openShutdownSignal()
```
thedevop, line 387: Not needed if part of NewClient.

server.go
```diff
@@ -383,6 +384,8 @@ func (s *Server) attachClient(cl *Client, listener string) error {
...
+	defer cl.sendShutdownSignal()
```
thedevop, line 388: Move this to the 1st line in attachClient.

thedevop:
I apologize this past week and the next few weeks I'll be busy on some other tasks. This looks good! We can simplify this a bit by always initialize shutdownSignal in newClient and always close shutdownSignal in attachClient.

server.go
```diff
@@ -1395,9 +1398,16 @@ func (s *Server) Close() error {
 // closeListenerClients closes all clients on the specified listener.
 func (s *Server) closeListenerClients(listener string) {
 	clients := s.Clients.GetByListener(listener)
+	var wg sync.WaitGroup
 	for _, cl := range clients {
-		_ = s.DisconnectClient(cl, packets.ErrServerShuttingDown)
-	}
+		wg.Add(1)
+		go func(c *Client) {
```
thedevop, line 1404: We may need to put some concurrent limits here, on a large server, this can be in the hundreds of thousands or even millions.

server.go
```diff
@@ -1395,9 +1398,16 @@ func (s *Server) Close() error {
...
+			s.DisconnectClient(c, packets.ErrServerShuttingDown)
+			c.waitForShutdownSignal()
+			wg.Done()
+		}(cl)
+	}
+	wg.Wait()
```
thedevop, line 1410: Add a max wait time?

thedevop:
I apologize this past week and the next few weeks I'll be busy on some other tasks. This looks good! We can simplify this a bit by always initialize shutdownSignal in newClient and always close shutdownSignal in attachClient.

werbenhu:
@thedevop NewClient() is called not only when creating new connections but also when called in server.loadClients() to load clients from the database. Clients loaded from the database won't have WriteLoop(). This is why I placed openShutdownSignal() inside attachClient() rather than in NewClient().

clients.go
werbenhu: NewClient() is called not only when creating new connections but also when called in server.loadClients() to load clients from the database. Clients loaded from the database won't have WriteLoop() and Read() loop. This is why I placed openShutdownSignal() inside attachClient() rather than in NewClient().

server.go
werbenhu: My idea here is that if cl.Read() is not executed, there is no need to initiate the signal.

server.go
werbenhu: Handling thousands of connections that need to be disconnected raises the question of how many Goroutines to limit. If the number of Goroutines is too low, it can lead to a situation where, after pressing Ctrl+C, the server takes a very long time to wait for all connections to disconnect.

werbenhu:
@thedevop @dgduncan @mochi-co When the server exits, handling the disconnection of all connections can be challenging, especially when dealing with a large number of connections. I'm wondering if a simpler approach could be used, such as not processing hook.OnDisconnect after using Ctrl+C.

server.go
thedevop: Few thousands (even tens of thousands) of Go routines should be fine. After all, for every connections, there are already 2 Go routines. It's just a matter how quickly these Go routines are spun up...  You can use [climit](https://github.com/thedevop1/climit) if you like.

server.go
thedevop: Make sense.

thedevop:
@werbenhu , for use cases that doesn't rely on OnDisconnect, #300 works. However, it may cause issue for use cases where that is relied upon. Note, in containerized environment, SIGTERM may occur more frequently.

server.go
werbenhu: Thanks, [climit](https://github.com/thedevop1/climit) looks very nice. I'll give it a try later.

clients.go
werbenhu: Refer to the previous comment.

clients.go
werbenhu: Refer to the previous comment.

server.go
werbenhu: Refer to the previous comment.

server.go
werbenhu: Done.

dgduncan:
@werbenhu @thedevop So I want to make sure I understand what is going on here. Basically there are multiple go routines running which is causing database to be closed before all clients have closed and all hooks have fired? So basically this is attempting to shut down all clients synchronously and then close all listeners?

werbenhu:
@dgduncan Your understanding is absolutely correct, and perhaps this is still worth discussing. I think #300 is also a good solution. A compromise could be to include this in the capabilities, allowing users to choose their preferred way to exit the server.

thedevop:
The issue is https://github.com/mochi-mqtt/server/blob/fac733fd71ebb292278b4b2b61f114fd3a058c44/server.go#L1385-L1389

s.Listeners.CloseAll(s.closeListenerClients) will close the connection, but the client is running in its own Go routine in attachClient. After it exit cl.Read, OnDisconnect hook will be called, and potentially OnUnsubscribed. Given these are executed in its own Go routine, Hook stop (L1388-1389) can be executed before these hooks.

@werbenhu added a "state" to indicate client has exited attachClient, and wait on it for all the clients before stop the hook.

server.go
```diff
@@ -1394,10 +1402,55 @@ func (s *Server) Close() error {
...
+	jobs := make(chan *Client, s.Options.Capabilities.ShutdownClientsWorkerNum)
+
+	// Create a context with a timeout. Adjust the timeout value as needed.
+	ctx, cancel := context.WithTimeout(context.Background(), time.Duration(s.Options.Capabilities.ShutdownClientsTimeout)*time.Second)
+	defer cancel()
+
+	// Start worker goroutines.
+	for i := 0; i < s.Options.Capabilities.ShutdownClientsWorkerNum; i++ {
+		wg.Add(1)
+		go func(clients <-chan *Client, ctx context.Context) {
```
dgduncan, line 1419: ctx should be first to keep standard context.Context pattern, unless this is a channel pattern that I am unaware of.

mochi-co:
I'm going to review as soon as possible, but have been a bit busy - apologies.

dgduncan (approved):
Seems like a necessary evil to have to do this. Approved!

server.go
werbenhu: @thedevop  I'm sorry for not using climit. climit is excellent, but the reason I didn't use it is that, when using climit, it didn't bring a significant improvement in code simplicity(there is also code for handling timeouts here).Instead, it concealed some details within climit. I extracted inspiration from climit, so thank you once again.

mochi-co:
I've had a bit of free time to study this issue, and my understanding is similar to what @thedevop and @werbenhu have suggested already, which is this:

1. `s.Close()` is called to initiate a graceful shutdown.
2. `s.Listeners.CloseAll(s.closeListenerClients`) issues shutdown signals to all listeners.
3. `s.Listeners.CloseAll` `l.wg.Wait()`'s until all listeners have disconnected all clients with `s.closeListenerClients` and returns.
4. `s.DisconnectClient` is called for each client, which issues a `ErrServerShuttingDown` code to each client. This packet is written to the client synchronously, before calling cl.Stop to end the blocking `err = cl.Read(s.receivePacket)` call in `attachClient`. 
5. This triggers the flow which calls `OnDisconnect`.
6. `s.hooks.OnStopped()` is called to trigger any OnStopped events.
7. `s.hooks.Stop()` is called to signal any running hooks to stop, and waits for all hooks to close down using `h.wg.Wait()` before returning.

This issue occurs because the a client can still be in the process of disconnecting (between `cl.Read` and `OnDisconnect`) when s.hooks.Stop() is called and issues a shutdown to all hooks. At a basic level, this is a race condition caused by the fact that the client disconnect (at a network level) happens in the client's attachClient goroutine.

Generally @werbenhu this is a great PR as always and it does appear to solve the problem.

That being said, I would like to offer up a perhaps simpler alternative solution to this issue - #301 . I would be interested if you could have a look at it and check if it fixes the issue.

PR #301, opened 2023-09-20
Refactor Listener WG to track clients

mochi-co:
Following #298 and #299, this PR presents a small solution to track the connection alive status of every client by repurposing the Listeners waitgroup to track clients instead of listeners. This change ensures that the Hooks shutdown sequence cannot be completed until all clients have successfully and sequentially exited their goroutines (between cl.Read and OnDisconnect). In this way, we can leverage the existing shutdown signally code.

Tests are unaffected, Paho passing.

### DIFF
--- a/listeners/listeners.go
+++ b/listeners/listeners.go
@@ -38,8 +38,8 @@ type Listener interface {
 
 // Listeners contains the network listeners for the broker.
 type Listeners struct {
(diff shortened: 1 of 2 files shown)


werbenhu:
@mochi-co  This code is absolutely perfect and effective, and it is also incredibly simple and clean, which is exactly what we all expect in code. Thanks.

werbenhu (approved):

thedevop (approved):

PR #303, opened 2023-09-23
Add some error logging in Listener.Serve().

werbenhu:
When I use an incorrect WebSocket address (for example, `go run main.go -ws=1882` where there is a missing colon before '1882'), it's challenging to pinpoint and discover where the problem lies. The server appears to be running fine. I suggest adding some error logging in Listener.Serve() to assist with troubleshooting.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -41,6 +41,7 @@ jobs:
 
   docker:
     runs-on: ubuntu-latest
+    if: github.repository == 'mochi-mqtt/server'
     needs: build
     steps:
       - name: Checkout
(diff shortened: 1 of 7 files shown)


dgduncan:
Should this be added to every listener? It seems like it would make sense that we bubble up the same error on every listener to make sure that the same issue does not effect the others.

werbenhu:
@dgduncan As it stands now, I believe we should, at the very least, add error logging within the Serve() methods of the WebSocket and HTTPStats listeners.

dgduncan (approved):

mochi-co (approved):
Looks good to me, disregarding the docker build errors as that's my fault 😄

Issue #293, opened 2023-09-07
paho golang client cannot keep stable connection


cyclamen:
I use the  paho golang client like the codes. The client prints:

2023/09/07 12:36:17 ERROR Connect lost: error=EOF
2023/09/07 12:36:17 INFO Connected
2023/09/07 12:36:27 ERROR Connect lost: error=EOF
2023/09/07 12:36:27 INFO Connected
2023/09/07 12:37:14 ERROR Connect lost: error=EOF
2023/09/07 12:37:14 INFO Connected
2023/09/07 12:37:22 ERROR Connect lost: error=EOF
2023/09/07 12:37:22 INFO Connected
2023/09/07 12:38:58 ERROR Connect lost: error=EOF
2023/09/07 12:38:58 INFO Connected
2023/09/07 12:39:23 ERROR Connect lost: error=EOF
2023/09/07 12:39:23 INFO Connected
2023/09/07 12:39:44 ERROR Connect lost: error=EOF
2023/09/07 12:39:44 INFO Connected
2023/09/07 12:42:52 ERROR Connect lost: error=EOF
2023/09/07 12:42:52 INFO Connected
2023/09/07 12:43:51 ERROR Connect lost: error=EOF
2023/09/07 12:43:51 INFO Connected
2023/09/07 12:43:59 ERROR Connect lost: error=EOF
2023/09/07 12:43:59 INFO Connected
.......

```
package main

import (
	"fmt"
	"github.com/eclipse/paho.mqtt.golang"
	"log/slog"
	"os"
	"time"
)

var f mqtt.MessageHandler = func(client mqtt.Client, msg mqtt.Message) {
	fmt.Printf("TOPIC: %s\n", msg.Topic())
	fmt.Printf("MSG: %s\n", msg.Payload())
}

func main() {
	//mqtt.DEBUG = log.New(os.Stdout, "", 0)
	//mqtt.ERROR = log.New(os.Stdout, "", 0)
	opts := mqtt.NewClientOptions().AddBroker("tcp://10.1.0.213:1883").SetClientID("abcdefghi1")
	opts.OnConnect = connectHandler
	opts.OnConnectionLost = connectLostHandler
	//opts.SetDefaultPublishHandler(messagePubHandler)
	opts.SetKeepAlive(2 * time.Second)
	opts.SetDefaultPublishHandler(f)
	opts.SetPingTimeout(1 * time.Second)

	c := mqtt.NewClient(opts)
[...]

PR #305, opened 2023-09-29
Emit warning if client keepalive is less than recommended minimum

mochi-co:
Following #293 and #296, this PR adds a warning if the client connecting has a keepalive set and the keepalive is below a minimum recommended keepalive value.

### DIFF
--- a/clients.go
+++ b/clients.go
@@ -8,6 +8,7 @@ import (
 	"bufio"
 	"bytes"
 	"context"
+	"errors"
 	"fmt"
 	"io"
 	"net"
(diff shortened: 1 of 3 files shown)


clients.go
```diff
@@ -21,8 +22,13 @@ import (
...
 
 const (
-	defaultKeepalive             uint16 = 10 // the default connection keepalive value in seconds
+	defaultKeepalive             uint16 = 10 // the default connection keepalive value in seconds.
 	defaultClientProtocolVersion byte   = 4  // the default mqtt protocol version of connecting clients (if somehow unspecified).
+	minimumKeepalive             uint16 = 5  // the minimum recommended keepalive - values under with display a warning.
+)
+
+var (
+	ErrMinimumKeepalive = errors.New("client keepalive is below minimum recommended value and may exhibit connection instability")
```
dgduncan, line 31: I think making this an error is good as it opens the door for a potential future in which the minimum keep alive can actually serve as an error and not just a warning.

dgduncan (approved):

werbenhu (approved):
This looks very good.

PR #307, opened 2023-09-30
Add a Chinese version of README.md.

werbenhu:
I took some time to translate the readme.md into Chinese and placed it in readme-cn.md. I hope this document will help more Chinese users, who may not be proficient in English, discover and use Mochi-server.


dgduncan (approved):
I love this. Allowing more people to gain access is always the right decision.

mochi-co (approved):
Great idea, totally agree. Approved.
Maybe it makes sense to have a indication of how up to date the translation is?

werbenhu:
@mochi-co This looks very necessary. If the README-CN.md needs updating next time, I will annotate the corresponding Release version at the beginning of the doc.

PR #315, opened 2023-10-04
Fix for unlimited maximum message expiry interval

### LINKED ISSUE #306: Default value of MaximumMessageExpiryInterval
We are having problems with retained messages being deleted exactly one day after publishing. I guess the source lies in the `server.Options.Capabilities.MaximumMessageExpiryInterval` property.

When starting the server with nil options (default capabilities):
```
server := mqtt.New(nil)
```
the `server.Options.Capabilities.MaximumMessageExpiryInterval` is set to 86400 seconds (exactly 1 day).

Shouldn't it be the maximum possible `int64` value to better represent the fact that the server was started without a maximum message expiry interval?

dadebue:
fixes #306


server_test.go
```diff
@@ -3276,6 +3279,36 @@ func TestServerClearExpiredRetained(t *testing.T) {
...
+	// No maximum expiry (math.MaxInt)
+	s = New(nil)
+	require.NotNil(t, s)
+
+	s.Options.Capabilities.MaximumMessageExpiryInterval = math.MaxInt
+
+	n = time.Now().Unix()
+	s.Topics.Retained.Add("a/b/c", packets.Packet{Created: n, Expiry: n - 1}) // explicit expiry set
+	s.Topics.Retained.Add("d/e/f", packets.Packet{Created: n})
+	s.Topics.Retained.Add("g/h/i", packets.Packet{Created: n - 99999999}) // kept even when old
```
werbenhu, line 3307: Is it better to use math.MaxInt to replace 99999999?

mochi-co:
I will try to look at this today, apologies for the delay

werbenhu (approved):

werbenhu (dismissed):

werbenhu (approved):

werbenhu (approved):

werbenhu (approved):

dgduncan (dismissed):
Looks like a simple and clean solution!

dadebue:
I still don't think that we should disable the clearing of expired retained messages if `s.Options.Capabilities.MaximumMessageExpi``ryInterval == math.MaxInt64`

What would be the result if a sender sends a retained message that has a expiry of 1 hour but s.Options.Capabilities.MaximumMessageExpiryInterval == math.MaxInt64? The message will not be deleted from the broker because clearing of retained messages is disabled and therefore will not expire in 1 hour as the sender intended...

What do you think?

werbenhu:
@dadebue Actually, according to the MQTT specification, retained messages do not provide the capability for client to set an expiration time. MQTT retained messages are a distinct message type that remains on the broker until they are replaced by a new retained message.

dadebue:
> @dadebue Actually, according to the MQTT spec, retained messages do not provide the capability for client to set an expiration time. MQTT retained messages remains on the broker until they are replaced by a new retained message or deleted when a new non-retained message is published.

OK I didn't know that. If that's the case, why do we even check for a expiry:
```
if (pk.Expiry > 0 && pk.Expiry < now) || ... {}
```

werbenhu:
I'm not entirely sure here; perhaps this is also a point worth discussing. @mochi-co @dgduncan @thedevop , does 'pk.Expiry' have the same effect on retained messages here? Should it be changed to something like this?
```go
// clearExpiredRetainedMessage deletes retained messages from topics if they have expired.
func (s *Server) clearExpiredRetainedMessages(now int64) {
	// If the maximum message expiry interval is set to math.MaxInt64, do not process expired messages.
	if s.Options.Capabilities.MaximumMessageExpiryInterval != math.MaxInt64 {
		for filter, pk := range s.Topics.Retained.GetAll() {
			if pk.Created < now-s.Options.Capabilities.MaximumMessageExpiryInterval {
				s.Topics.Retained.Delete(filter)
				s.hooks.OnRetainedExpired(filter)
			}
		}
	}
}```

**Comment by dgduncan** [comment]
@werbenhu @dadebue I can not find where it is precisely stated; however, according to [3.3.1.3 RETAIN](https://docs.oasis-open.org/mqtt/mqtt/v5.0/os/mqtt-v5.0-os.html#_Toc3901104) 

> If the current retained message for a Topic expires, it is discarded and there will be no retained message for that topic.

And according to the [HiveMQ MQTT 5 Essentials](https://www.hivemq.com/blog/mqtt5-essentials-part4-session-and-message-expiry/)

> Furthermore, if the ‘retained=true’ option is selected during the PUBLISH message, the interval also dictates the length of time a message is retained on a particular topic.

**Comment by dadebue** [comment]
And even more in the mentioned HiveMQ essential article:

> For retained messages, the message expiry interval operates similarly, guaranteeing that these messages are only dispatched to new subscribers for a specified period.

Sounds to me as retained messages can have an expiry interval just as regular messages...

**Comment by mochi-co** [comment]
> And even more in the mentioned HiveMQ essentials article:
> 
> > For retained messages, the message expiry interval operates similarly, guaranteeing that these messages are only dispatched to new subscribers for a specified period.
> 
> Sounds to me as retained messages can have an expiry interval just as regular messages...

I'm planning to properly re-read through and address all of these comments (and others) as soon as I can, but the above is exactly correct, as far as I remember. At least for MQTTv5, [the client may specify how long the message may persist](https://docs.oasis-open.org/mqtt/mqtt/v5.0/os/mqtt-v5.0-os.html#_Toc3901112). As MQTTv3 doesn't have client-set message expiries, we fall back to the server capabilities instead (to avoid retaining all messages forever by default).

**Comment by werbenhu** [comment]
If, according to the specification, there isn't a very clear indication of whether the message expiration time is effective for retained messages, perhaps the broker can make a non-normative decision on its own about this. 

Personally, I lean towards the idea that message expiration should be effective for retained messages as well.

Keeping retained messages indefinitely on the broker could be problematic, and giving clients the ability to determine how long retained messages should be kept seems more reasonable.

**Comment by mochi-co** [comment]
I think it's more of a lack of clarity in the specification, but it is there:

https://docs.oasis-open.org/mqtt/mqtt/v5.0/os/mqtt-v5.0-os.html#_Toc3901112

> If the Message Expiry Interval has passed and the Server has not managed to start onward delivery to a matching subscriber, then it MUST delete the copy of the message for that subscriber

This applies to all messages, including retained messages and will messages: they must be deleted if they have not started onward delivery to a subscriber by the time the expiry period has passed.

**Comment by werbenhu** [comment]
> I still think that we shouldn't disable the clearing of expired retained messages if `` s.Options.Capabilities.MaximumMessageExpi``ryInterval == math.MaxInt64 ``
> 
> What would be the result if a sender sends a retained message that has a expiry of 1 hour but `s.Options.Capabilities.MaximumMessageExpiryInterval == math.MaxInt64`? The message will not be deleted from the broker because clearing of retained messages is disabled and therefore will not expire in 1 hour as the sender intended...
> 
> What do you think?

@dadebue  you can now proceed to submit code changes based on this idea. @mochi-co  it appears that we've lost the permission to push and submit to the forked branch.

**Comment by mochi-co** [comment]
I'm just catching up on this but to confirm:

- Setting `s.Options.Capabilities.MaximumMessageExpiryInterval == math.MaxInt64` should not disable message expiry, it should be honoured even though it is infinitely in the future - unless there is a significant performance reason not to do so (I am unaware of such a thing currently).
- Even if `s.Options.Capabilities.MaximumMessageExpiryInterval` is `0`, message expiry _per message_ must still be honoured per the MQTTv5 spec if set in the message properties, but MQTTv3 messages and those without expiries should be retained forever.

File: clients.go [hunk 2]
```diff
@@ -330,7 +330,7 @@ func (cl *Client) ResendInflightMessages(force bool) error {
 func (cl *Client) ClearInflights(now, maximumExpiry int64) []uint16 {
 	deleted := []uint16{}
 	for _, tk := range cl.State.Inflight.GetAll(false) {
-		if (tk.Expiry > 0 && tk.Expiry < now) || tk.Created+maximumExpiry < now {
+		if (tk.Expiry > 0 && tk.Expiry < now) || tk.Created < now-maximumExpiry {
```
mochi-co, line 333: This addresses the issue in #306 but I feel that this change may be harder to understand without context. Also, if maximum expiry is maxInt, we don't need to evaluate this on every iteration. 

I'll look at this whole issue more deeply tomorrow as I should have some time 👍🏻

server.go
```diff
@@ -1596,20 +1596,26 @@ func (s *Server) clearExpiredClients(dt int64) {
 
 // clearExpiredRetainedMessage deletes retained messages from topics if they have expired.
 func (s *Server) clearExpiredRetainedMessages(now int64) {
-	for filter, pk := range s.Topics.Retained.GetAll() {
-		if (pk.Expiry > 0 && pk.Expiry < now) || pk.Created+s.Options.Capabilities.MaximumMessageExpiryInterval < now {
-			s.Topics.Retained.Delete(filter)
-			s.hooks.OnRetainedExpired(filter)
+	// If the maximum message expiry interval is set to math.MaxInt64, do not process expired messages.
+	if s.Options.Capabilities.MaximumMessageExpiryInterval != math.MaxInt64 {
```
mochi-co, line 1600: If we were to implement this change, and I'm not certain we should, this may be more in line with the existing style:

```suggestion
	if s.Options.Capabilities.MaximumMessageExpiryInterval == math.MaxInt64 {
	    return
	}
	
	for filter, pk := range s.Topics.Retained.GetAll() {
	
```

mochi-co (changes requested):
I gave the code a quick review (it's 2am here but I wanted to get an understanding of what's happening), I feel that the intention is correct, but there's some optimisations we could make for readability and performance. Also see my notes above about spec conformity, which we may be deviating from with part of this change.

mochi-co:
> @mochi-co it appears that we've lost the permission to push and submit to the forked branch.

This may be related to the permissions changes I made. I will investigate 👍🏻

clients.go
werbenhu: @mochi-co Using addition here might lead to integer overflow, while subtraction can prevent it.

mochi-co: @werbenhu I agree with you, however I just want to make sure we have the most considered solution as it was not clear to me when I first read it

werbenhu: You're right; perhaps we can use another, clearer approach to address the overflow issue.

mochi-co:
@werbenhu @dadebue @dgduncan I had a little look at this, and I think this is the approach I would take if I were trying to fix the `MaxInt` issue described. It is a bit more 'wordy' than the previous changes, but I feel it might be a bit easier to understand were we to come back to this code in 5 years and wonder what it does. All credit goes to @werbenhu however for identifying the issue in the first place!

```diff
diff --git forkSrcPrefix/server.go forkDstPrefix/server.go
index 8b7b46da77734338be58126635a96576dd270909..bb343cceaab7eafafb15a4b9d17710502ccde346 100644
--- forkSrcPrefix/server.go
+++ forkDstPrefix/server.go
@@ -1597,7 +1597,11 @@ func (s *Server) clearExpiredClients(dt int64) {
 // clearExpiredRetainedMessage deletes retained messages from topics if they have expired.
 func (s *Server) clearExpiredRetainedMessages(now int64) {
 	for filter, pk := range s.Topics.Retained.GetAll() {
-		if (pk.Expiry > 0 && pk.Expiry < now) || pk.Created+s.Options.Capabilities.MaximumMessageExpiryInterval < now {
+		serverExpiry := pk.Created + s.Options.Capabilities.MaximumMessageExpiryInterval
+		if serverExpiry < 0 { // catch int overflows if pk.Created + MaximumMessageExpiryInterval > math.MaxInt
+			serverExpiry = s.Options.Capabilities.MaximumMessageExpiryInterval
+		}
+		if (pk.Expiry > 0 && pk.Expiry < now) || pk.Created > serverExpiry {
 			s.Topics.Retained.Delete(filter)
 			s.hooks.OnRetainedExpired(filter)
 		}
diff --git forkSrcPrefix/server_test.go forkDstPrefix/server_test.go
index ec0dc3b6ea4d9e58166f5ee0e2d436baf38077b0..e27d3726d4271b04a35c3c9b940bfdd1bb1f77dc 100644
--- forkSrcPrefix/server_test.go
+++ forkDstPrefix/server_test.go
@@ -9,6 +9,7 @@ import (
 	"encoding/binary"
 	"io"
 	"log/slog"
+	"math"
 	"net"
 	"strconv"
 	"sync"
@@ -3278,6 +3279,23 @@ func TestServerClearExpiredRetained(t *testing.T) {
 	require.Len(t, s.Topics.Retained.GetAll(), 2)
 }
 
+func TestServerClearExpiredRetainedMaxInt(t *testing.T) {
+	s := New(nil)
+	require.NotNil(t, s)
+	s.Options.Capabilities.MaximumMessageExpiryInterval = math.MaxInt
+
+	n := time.Now().Unix()
+	s.Topics.Retained.Add("a/b/c", packets.Packet{Created: n, Expiry: n - 1})
+	s.Topics.Retained.Add("d/e/f", packets.Packet{Created: n, Expiry: n - 2})
+	s.Topics.Retained.Add("g/h/i", packets.Packet{Created: n - 3}) // within bounds
+	s.Topics.Retained.Add("j/k/l", packets.Packet{Created: n - 5})
+	s.Topics.Retained.Add("m/n/o", packets.Packet{Created: n})
+
+	require.Len(t, s.Topics.Retained.GetAll(), 5)
+	s.clearExpiredRetainedMessages(n)
+	require.Len(t, s.Topics.Retained.GetAll(), 3)
+}
+
 func TestServerClearExpiredClients(t *testing.T) {
 	s := New(nil)
 	require.NotNil(t, s)
diff --git forkSrcPrefix/clients.go forkDstPrefix/clients.go
index 6d5ff9a0a8d416ab0036faabcab5fa822d6c136d..cd975bf39fa6b32388cc277f64e139ccba16aecb 100644
--- forkSrcPrefix/clients.go
+++ forkDstPrefix/clients.go
@@ -330,7 +330,11 @@ func (cl *Client) ResendInflightMessages(force bool) error {
 func (cl *Client) ClearInflights(now, maximumExpiry int64) []uint16 {
 	deleted := []uint16{}
 	for _, tk := range cl.State.Inflight.GetAll(false) {
-		if (tk.Expiry > 0 && tk.Expiry < now) || tk.Created+maximumExpiry < now {
+		serverExpiry := tk.Created + maximumExpiry
+		if serverExpiry < 0 { // catch int overflows if pk.Created + maximumExpiry > math.MaxInt
+			serverExpiry = maximumExpiry
+		}
+		if (tk.Expiry > 0 && tk.Expiry < now) || tk.Created > serverExpiry {
 			if ok := cl.State.Inflight.Delete(tk.PacketID); ok {
 				cl.ops.hooks.OnQosDropped(cl, tk)
 				atomic.AddInt64(&cl.ops.info.Inflight, -1)
diff --git forkSrcPrefix/clients_test.go forkDstPrefix/clients_test.go
index e992b5dc718cee6b882f5ad3a773cf929106eec1..b6ea4e6e508f5c8d6d0ed784fd8b41427a03de0c 100644
--- forkSrcPrefix/clients_test.go
+++ forkDstPrefix/clients_test.go
@@ -11,6 +11,7 @@ import (
 	"errors"
 	"io"
 	"log/slog"
+	"math"
 	"net"
 	"strings"
 	"sync/atomic"
@@ -317,6 +318,23 @@ func TestClientClearInflights(t *testing.T) {
 	require.Equal(t, 2, cl.State.Inflight.Len())
 }
 
+func TestClientClearInflightsMaxInt(t *testing.T) {
+	cl, _, _ := newTestClient()
+
+	n := time.Now().Unix()
+	cl.State.Inflight.Set(packets.Packet{PacketID: 1, Expiry: n - 1})
+	cl.State.Inflight.Set(packets.Packet{PacketID: 2, Expiry: n - 2})
+	cl.State.Inflight.Set(packets.Packet{PacketID: 3, Created: n - 3}) // within bounds
+	cl.State.Inflight.Set(packets.Packet{PacketID: 5, Created: n - 5})
+	cl.State.Inflight.Set(packets.Packet{PacketID: 7, Created: n})
+	require.Equal(t, 5, cl.State.Inflight.Len())
+
+	deleted := cl.ClearInflights(n, math.MaxInt)
+	require.Len(t, deleted, 2)
+	require.ElementsMatch(t, []uint16{1, 2}, deleted)
+	require.Equal(t, 3, cl.State.Inflight.Len())
+}
+
 func TestClientResendInflightMessages(t *testing.T) {
 	pk1 := packets.TPacketData[packets.Puback].Get(packets.TPuback)
 	cl, r, w := newTestClient()

```

werbenhu:
@dadebue @mochi-co can you make changes to the code and submit it to resolve the unresolved conversations here? @mochi-co I still don't have the permission to modify the code when I review it, which also means I don't have the permission to submit code to the forked branch.

mochi-co:
@werbenhu Try now

werbenhu (dismissed):

mochi-co:
@werbenhu The current changes on the branch don't fully comply with the scenarios in https://github.com/mochi-mqtt/server/pull/315#issuecomment-1762419335 - With `s.Options.Capabilities.MaximumMessageExpiryInterval == math.MaxInt64` we should still continue to expire messages which have their own set expiry values set. 

I'd recommend reviewing the diff I posted above as I believe this addresses all the scenarios we are trying to fix.

werbenhu:
> I'm just catching up on this and it sounds like there are some misunderstandings on the spec and current design, but to confirm:
> 
> * Setting `s.Options.Capabilities.MaximumMessageExpiryInterval == math.MaxInt64` should not disable message expiry, it should be honoured even though it is infinitely in the future - unless there is a significant performance reason not to do so (I am unaware of such a thing currently).
> * if `s.Options.Capabilities.MaximumMessageExpiryInterval` is `1` or more, message expiry _per message_ must still be honoured per the MQTTv5 spec if set in the message properties, but MQTTv3 messages and those without expiries should be retained forever.
> * if `s.Options.Capabilities.MaximumMessageExpiryInterval` is `0`, message expiry _per message_ must be ignored for all messages and protocols.
> 
> (edited - apologies it's late and I've been extraordinarily busy)

It's done now.

werbenhu:
There still seem to be some issues here, I'll need to make some submissions later

clients.go
```diff
@@ -329,8 +329,12 @@ func (cl *Client) ResendInflightMessages(force bool) error {
 // ClearInflights deletes all inflight messages for the client, e.g. for a disconnected user with a clean session.
 func (cl *Client) ClearInflights(now, maximumExpiry int64) []uint16 {
 	deleted := []uint16{}
+
 	for _, tk := range cl.State.Inflight.GetAll(false) {
-		if (tk.Expiry > 0 && tk.Expiry < now) || tk.Created+maximumExpiry < now {
+		expired := tk.ProtocolVersion == 5 && (tk.Expiry > 0 && tk.Expiry < now) // [MQTT-3.3.2-5]
```
mochi-co, line 334:
```suggestion
		expired := tk.ProtocolVersion == 5 && tk.Expiry > 0 && tk.Expiry < now // [MQTT-3.3.2-5]
```

server.go
```diff
@@ -1596,8 +1596,17 @@ func (s *Server) clearExpiredClients(dt int64) {
...
 func (s *Server) clearExpiredRetainedMessages(now int64) {
+
+	// If the maximum message expiry interval is set to 0, do not process expired messages for all protocols.
+	if s.Options.Capabilities.MaximumMessageExpiryInterval == 0 {
+		return
+	}
+
 	for filter, pk := range s.Topics.Retained.GetAll() {
-		if (pk.Expiry > 0 && pk.Expiry < now) || pk.Created+s.Options.Capabilities.MaximumMessageExpiryInterval < now {
+		expired := pk.ProtocolVersion == 5 && (pk.Expiry > 0 && pk.Expiry < now) // [MQTT-3.3.2-5]
```
mochi-co, line 1606:
```suggestion
		expired := pk.ProtocolVersion == 5 && pk.Expiry > 0 && pk.Expiry < now // [MQTT-3.3.2-5]
```

mochi-co:
> There still seem to be some issues here, I'll need to make some submissions later

Which issues are you seeing?

werbenhu:
> I still think that we shouldn't disable the clearing of expired retained messages if `` s.Options.Capabilities.MaximumMessageExpi``ryInterval == math.MaxInt64 ``
> 
> What would be the result if a sender sends a retained message that has a expiry of 1 hour but `s.Options.Capabilities.MaximumMessageExpiryInterval == math.MaxInt64`? The message will not be deleted from the broker because clearing of retained messages is disabled and therefore will not expire in 1 hour as the sender intended...
> 
> What do you think?

@mochi-co I think @dadebue is right, and I've made some changes to it.

werbenhu:
@mochi-co take a look at the current implementation. I always feel that what's written here in `func (cl *Client) ClearInflights(now, maximumExpiry int64) []uint16` isn't concise and clear enough;

 'maximumExpiry=0' carries a dual meaning:

- When calling cl.ClearInflights() within server.clearExpiredInflights, 'maximumExpiry=0' means there's no need to process the maximum.
- When calling cl.ClearInflights() within server.inheritClientSession, 'maximumExpiry=0' indicates the need to forcibly clear all client inflights.

dadebue:
I like the current implementation.

Maybe we should change the name `abandoned` to something different. What the maximum expiry interval is doing is forcing the expiry of old messages. Why don't we call it `forceExpiry` then?

werbenhu:
@dadebue @mochi-co   I have added `client.ClearExpiredInflights()` to clear expired messages, while `client.ClearInflights()` is used to clear all inflight messages. Please review the code again. Thanks.

werbenhu (approved):

dadebue:
I approve the changes with this comment ☺️ (I can't approve in GitHub as I wasn't requested to review)

werbenhu:
@dadebue Because you are the author of this PR.:smile:

mochi-co (approved):
It took a bit of a confusing road to get here, but I've given it a thorough review and it makes sense to me, and I think this is an excellent solution which makes the overall codebase a bit cleaner and easier to understand. Paho and unit tests are also passing, so I approve! Great job @werbenhu @dadebue

dadebue:
@mochi-co @dgduncan @werbenhu what's is the release schedule for this repo? When will all the changes since October 2nd be released?

mochi-co:
@dadebue Apologies, I've had some big life stuff going on and the time got away from me. I've released these changes as v2.4.2 :)

dadebue:
No worries and thanks :)

PR #321, opened 2023-10-20
improve transport performance with bufio

x20080406:
This code snippet can improve the network throughput for the subscriber.


mochi-co (changes requested):
I love anything which can increase throughput, however there are some significant test failures which need to be addressed before we can properly review and merge 🙂

x20080406:
> I love anything which can increase throughput, however there are some significant test failures which need to be addressed before we can properly review and merge 🙂

Yes. I've fixed them. Root cause: the variable Conn in ClientConnection has been reset in server_test.go, but the bconn does not been reset.

mochi-co:
Looks good to me, this will go out in the next release - thank you @x20080406 !

PR #333, opened 2023-11-24
Demo how to subscribe and publish directly within examples/hooks.

werbenhu:
In issues #332  and #330 , users raised the same question about how to use subscribe and publish directly within a hook. Therefore, I added a demonstration in examples/hooks on how to achieve this.


mochi-co (approved):
Looks good to me, thank you :)

PR #336, opened 2023-11-24
feat: return tcp.Address from listener, if exists

AeroNotix:
* This is the more accurate and correct address of the listener
* Useful if you want to listen on port 0 to dynamically create listeners (think of unit/integration tests)

### DIFF
--- a/listeners/tcp.go
+++ b/listeners/tcp.go
@@ -44,6 +44,9 @@ func (l *TCP) ID() string {
 
 // Address returns the address of the listener.
 func (l *TCP) Address() string {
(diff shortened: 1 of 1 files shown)


AeroNotix:
This should always match what the original data was _unless_ port 0 was selected. Don't think there should be any risk?

mochi-co:
@AeroNotix As nobody else has any concerns, and I agree with your assessment, so it's getting merged. I think this is a great idea. Thank you very much!

AeroNotix:
Thanks!

Issue #339, opened 2023-12-06
invalid client status which loaded from storage.


x20080406:
Hi, there
     When invoke the the method `client.Closed()` with client loaded from storage, but the client not reconnect to server . it still  return false.  I think there should invoke the `client.Stop` at the end of `server.loadClients`

```go
// loadClients restores clients from the datastore.
func (s *Server) loadClients(v []storage.Client) {
	for _, c := range v {
		cl := s.NewClient(nil, c.Listener, c.ID, false)
		cl.Properties.Username = c.Username
		cl.Properties.Clean = c.Clean
		cl.Properties.ProtocolVersion = c.ProtocolVersion
		cl.Properties.Props = packets.Properties{
			SessionExpiryInterval:     c.Properties.SessionExpiryInterval,
			SessionExpiryIntervalFlag: c.Properties.SessionExpiryIntervalFlag,
			AuthenticationMethod:      c.Properties.AuthenticationMethod,
			AuthenticationData:        c.Properties.AuthenticationData,
			RequestProblemInfoFlag:    c.Properties.RequestProblemInfoFlag,
			RequestProblemInfo:        c.Properties.RequestProblemInfo,
			RequestResponseInfo:       c.Properties.RequestResponseInfo,
			ReceiveMaximum:            c.Properties.ReceiveMaximum,
			TopicAliasMaximum:         c.Properties.TopicAliasMaximum,
			User:                      c.Properties.User,
			MaximumPacketSize:         c.Properties.MaximumPacketSize,
		}
		cl.Properties.Will = Will(c.Will)
                cl.Stop(packets.ErrServerShuttingDown) // make it stopped
		s.Clients.Add(cl)		
	}
}
```

Another option. I think there should add a `connected` int64 for `ClientState`.

Thank you.

thedevop:
@x20080406 , good find! It is a bug, as a new context is created with newClient, but it is not canceled as there is no disconnect activity. What you proposed would work, I would move it up to immediately after creating the new client though.
```go
		cl := s.NewClient(nil, c.Listener, c.ID, false)
		cl.Stop()
```

werbenhu:
@thedevop I think moving `go cl.WriteLoop()` out of NewClient() and placing it in either server.EstablishConnection() or server.attachClient() would be more reasonable.

thedevop:
We can add a new condition check that if c net.Conn is nil, don't start the WriteLoop.

thedevop:
[...]

PR #341, opened 2023-12-07
Handle expired clients in server.loadClients().

werbenhu:
I encountered a problem similar to #339 . If the broker is forcefully shut down, and clients have the clean flag set, after restarting the broker and calling loadClients(), these expired clients will be retained indefinitely if the client is no longer connected.

It is necessary to handle expired clients in server.loadClients(). Call OnDisconnect() to trigger the hook for clearing persistent clients. Also, if a client has expired, there is no need to retain this client in server.Clients.

### DIFF
--- a/server.go
+++ b/server.go
@@ -1559,7 +1559,6 @@ func (s *Server) loadClients(v []storage.Client) {
 		if expire {
 			cl.ClearInflights()
 			s.UnsubscribeClient(cl)
-			s.Clients.Delete(cl.ID)
 		} else {
 			s.Clients.Add(cl)
 		}
--- a/server.go
+++ b/server.go
@@ -1553,7 +1553,16 @@ func (s *Server) loadClients(v []storage.Client) {
 			MaximumPacketSize:         c.Properties.MaximumPacketSize,
 		}
 		cl.Properties.Will = Will(c.Will)
-		s.Clients.Add(cl)
+
+		expire := (cl.Properties.ProtocolVersion == 5 && cl.Properties.Props.SessionExpiryInterval == 0) || (cl.Properties.ProtocolVersion < 5 && cl.Properties.Clean)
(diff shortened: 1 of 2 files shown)


mochi-co (approved):
Makes sense to me, and a good solution. For me, it's approved! Thank you for finding this @werbenhu!

x20080406:
I think this change make the logic become too difficult. The loaded client should be invoke the `Stop` method only. 

With this commit, there will get some issues.
```
expire := (cl.Properties.ProtocolVersion == 5 && cl.Properties.Props.SessionExpiryInterval == 0) || (cl.Properties.ProtocolVersion < 5 && cl.Properties.Clean)
s.hooks.OnDisconnect(cl, packets.ErrServerShuttingDown, expire)
if expire {
	cl.ClearInflights()  // inflights not recoverd
	s.UnsubscribeClient(cl)  // subscription not recoverd
        // there couldn't find clients after loadClients in the `readStore` method,
} else {
	s.Clients.Add(cl)
}
```

werbenhu:
@x20080406 Now that 'cl' has already been deprecated, and you have obtained all the information from it, what is the significance of keeping it continued in 'Clients'?

mochi-co:
Released in v2.4.3
This is a preventative fix to solve the current issue as it's better than having bugs, however I am happy for @x20080406 and @werbenhu for you to please keep discussing the optimal solution to replace this one if necessary.

thedevop:
@x20080406 @werbenhu @mochi-co 

s.NewClient will start the WriteLoop in a Go routine for non-inline clients:
https://github.com/mochi-mqtt/server/blob/624dde0986e28442a1406145de637c835190b10d/server.go#L231-L239
This will only stop if cl.Stop or cl.State.cancelOpen is called.

werbenhu:
@thedevop I found this too, and I will submit a new pull request to fix it.

Issue #342, opened 2023-12-13
Packet encoding optimization


thedevop:
@mochi-co, I found some optimization opportunities when encoding packets:
1. During PublishEncode, write pk.Payload only once instead of twice
2. Use a memory pool (sync.Pool) for Buffer used during encodings

Lastly, in pk.Properties.Encode, a new Buffer is allocated:
https://github.com/mochi-mqtt/server/blob/624dde0986e28442a1406145de637c835190b10d/packets/properties.go#L197-L202

However in all instances where this function is called, a new Buffer is already allocated specifically for this: 
https://github.com/mochi-mqtt/server/blob/624dde0986e28442a1406145de637c835190b10d/packets/packets.go#L612-L616

Should we remove the NewBuffer within pk.Properties.Encode and use the Buffer that is passed in. This will also rsult in one less write of the property.

mochi-co:
Released in v2.4.4 🚀

PR #343, opened 2023-12-13
Packet encoding optimization

thedevop:
As described in #342.

### DIFF
--- a/mempool/bufpool.go
+++ b/mempool/bufpool.go
@@ -1,0 +1,81 @@
+package mempool
+
+import (
+	"bytes"
+	"sync"
+)
+
+var bufPool = NewBuffer(64 * 1024)
+
+// GetBuffer takes a Buffer from the default buffer pool
+func GetBuffer() *bytes.Buffer { return bufPool.Get() }
+
+// PutBuffer returns Buffer to the default buffer pool
+func PutBuffer(x *bytes.Buffer) { bufPool.Put(x) }
+
+type BufferPool interface {
+	Get() *bytes.Buffer
+	Put(x *bytes.Buffer)
+}
+
+// NewBuffer returns a buffer pool. The max specify the max capacity of the Buffer the pool will
+// return. If the Buffer becoomes large than max, it will no longer be returned to the pool. If
+// max <= 0, no limit will be enforced.
+func NewBuffer(max int) BufferPool {
+	if max > 0 {
+		return newBufferWithCap(max)
+	}
+
+	return newBuffer()
+}
+
+// Buffer is a Buffer pool.
+type Buffer struct {
+	pool *sync.Pool
+}
+
+func newBuffer() *Buffer {
+	return &Buffer{
+		pool: &sync.Pool{
+			New: func() any { return new(bytes.Buffer) },
+		},
+	}
+}
+
+// Get a Buffer from the pool.
+func (b *Buffer) Get() *bytes.Buffer {
+	return b.pool.Get().(*bytes.Buffer)
+}
+
+// Put the Buffer back into pool. It resets the Buffer for reuse.
+func (b *Buffer) Put(x *bytes.Buffer) {
+	x.Reset()
+	b.pool.Put(x)
+}
+
+// BufferWithCap is a Buffer pool that
+type BufferWithCap struct {
+	bp  *Buffer
+	max int
+}
+
+func newBufferWithCap(max int) *BufferWithCap {
+	return &BufferWithCap{
+		bp:  newBuffer(),
+		max: max,
+	}
+}
+
+// Get a Buffer from the pool.
+func (b *BufferWithCap) Get() *bytes.Buffer {
+	return b.bp.Get()
+}
+
+// Put the Buffer back into the pool if the capacity doesn't exceed the limit. It resets the Buffer
+// for reuse.
+func (b *BufferWithCap) Put(x *bytes.Buffer) {
+	if x.Cap() > b.max {
+		return
+	}
+	b.bp.Put(x)
+}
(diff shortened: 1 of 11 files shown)


mochi-co:
I love the idea -how do we feel about the risk of active items in the pool being dropped under high load situations? Is this risk present?

thedevop:
No active items will be in the pool. Currently a buffer is 1st taken out of the pool, if none is available, a new buffer is generated. It will only be put back into the pool when it's no longer needed.

Any items in the pool can be GCed any time, which will have no negative impact.

thedevop:
@werbenhu, I have checked in benchmark, the results are:
```BenchmarkBufferPool
cpu: Intel(R) Core(TM) i7-7920HQ CPU @ 3.10GHz
BenchmarkBufferPool
BenchmarkBufferPool-8                	52733648	        20.26 ns/op	       0 B/op	       0 allocs/op
BenchmarkBufferPoolWithCapLarger
BenchmarkBufferPoolWithCapLarger-8   	56555937	        21.71 ns/op	       0 B/op	       0 allocs/op
BenchmarkBufferPoolWithCapLesser
BenchmarkBufferPoolWithCapLesser-8   	 6784573	       182.1 ns/op	     112 B/op	       2 allocs/op
BenchmarkBufferWithoutPool
BenchmarkBufferWithoutPool-8         	26074543	        49.38 ns/op	      64 B/op	       1 allocs/op
```
Actual performance should be even better, as the bytes.Buffer may grow few times to encode the packet.

I used to use the simplified method as you have provided, which works perfectly fine if all the items are similar in size. It doesn't work as well if items are of different size. See https://golang.org/issue/23199. Although most of the packets without payload should be similar in size, but with the inclusion of metadata (user properties), it may become quite large.

I intentionally for this version avoid the final buffer that includes the payload, we should use a dedicated buffer pool for that, and perhaps let it be user configurable based on their use case.

thedevop:
I benchmarked pk.Properties.Encode using the following:

```go
func BenchmarkEncodeNoPool(b *testing.B) {
	props := propertiesStruct

	for i := 0; i < b.N; i++ {
		buf := bytes.NewBuffer([]byte{})
		props.Encode(Reserved, Mods{AllowResponseInfo: true}, buf, 0)
		_ = buf
	}
}

func BenchmarkEncodePoolBuf(b *testing.B) {
	props := propertiesStruct

	for i := 0; i < b.N; i++ {
		buf := mempool.GetBuffer()
		props.Encode(Reserved, Mods{AllowResponseInfo: true}, buf, 0)
		mempool.PutBuffer(buf)
	}
}
```

The results are:
```
cpu: Intel(R) Core(TM) i7-7920HQ CPU @ 3.10GHz
BenchmarkEncodeNoPool-8    	  664742	      1716 ns/op	     736 B/op	       9 allocs/op
BenchmarkEncodePoolBuf-8   	  748135	      1574 ns/op	     504 B/op	       6 allocs/op
```

If I also replace the buf within Encode to use buffer pool, the results are:
```
cpu: Intel(R) Core(TM) i7-7920HQ CPU @ 3.10GHz
BenchmarkEncodeNoPool-8    	  730120	      1552 ns/op	     288 B/op	       6 allocs/op
BenchmarkEncodePoolBuf-8   	  873458	      1352 ns/op	      56 B/op	       3 allocs/op
```

mempool/bufpool.go
```diff
@@ -0,0 +1,81 @@
...
+
+// Get a Buffer from the pool.
+func (b *BufferWithCap) Get() *bytes.Buffer {
+	return b.bp.Get()
+}
+
+// Put the Buffer back into the pool if the capacity doesn't exceed the limit. It resets the Buffer
+// for reuse.
+func (b *BufferWithCap) Put(x *bytes.Buffer) {
+	if x.Cap() > b.max {
```
werbenhu, line 77: Perhaps adding some comments here would be more beneficial for understanding the code, such as 
```go
// Proper usage of a sync.Pool requires each entry to have approximately
// the same memory cost. To obtain this property when the stored type
// contains a variably-sized buffer, we add a hard limit on the maximum buffer
// to place back in the pool.
//
// See https://golang.org/issue/23199
// Also refer to https://go-review.googlesource.com/c/go/+/136035
if x.Cap() > b.max {
```

werbenhu:
@thedevop @mochi-co Can we consider retaining only the usage of BufferWithCap and eliminating the approach of using Buffer Poll without a cap? 

See https://go-review.googlesource.com/c/go/+/136116 and https://go-review.googlesource.com/c/go/+/136035 for reference.

werbenhu:
@thedevop perhaps we can learn from [encoding/json: fix usage of sync.Pool](https://go-review.googlesource.com/c/go/+/136035), where we can see that the standard library's json.Marshal also faces a similar issue to what we are currently encountering. However, at least for now, it appears that the json.Marshal still directly employs the sync.Pool without a cap. Personally, I suggest we can temporarily postpone the use of BufferWithCap. @mochi-co  what are your thoughts on this issue?

thedevop:
@werbenhu, I'm ok we use the pool without cap, as most of the packet encoding without payload should vary too much on size. I updated the default pool to not have a cap.

werbenhu (approved):

mochi-co:
Sorry my life has been crazy this year. Just catching up on all this good work you've done.

The pool benchmarks are promising, and at scale this will make a significant difference. @thedevop your point about differing payload sizes is poignant and well considered.

I agree with @werbenhu that we should first attempt to use the pool without a cap to reduce the complexity we're adding here. However, we should monitor the change and add the cap if it does become necessary.

As for me, I think this is a great contribution, and it's definitely getting merged.

mochi-co (approved):

PR #344, opened 2023-12-13
Move cl.WriteLoop() to attachClient() and call cl.Stop() in loadClients() to update client.State.

werbenhu:
In the server.loadClients() and when using inlineClient, NewClient() is called, but these may not necessarily represent actual connections. Calling go cl.WriteLoop() within NewClient() leads to a goroutine leak. I think moving go cl.WriteLoop() out of NewClient() and placing it in either server.EstablishConnection() or server.attachClient() would be more reasonable.


thedevop (approved):

werbenhu:
@thedevop @x20080406 @mochi-co @dgduncan I have added a call to cl.Stop() in loadClients() to cancel the client's context, update cl.State with information such as disconnected time, and set the stopCause.

Please let me know if there are any other cases that need to be considered.

thedevop (approved):

x20080406 (approved):

mochi-co (approved):
This is a much better idea!

Issue #346, opened 2023-12-16
Data race in buffer with v2.4.3


jkager:
I keep getting data races with v2.4.3 while running tests in GitHub workflows:

```
WARNING: DATA RACE
Write at 0x00c00137f0e0 by goroutine 1116:
  bytes.(*Buffer).grow()
      /opt/hostedtoolcache/go/1.21.4/x64/src/bytes/buffer.go:151 +0x2d3
  bytes.(*Buffer).Write()
      /opt/hostedtoolcache/go/1.21.4/x64/src/bytes/buffer.go:179 +0xc4
  bytes.(*Buffer).WriteTo()
      /opt/hostedtoolcache/go/1.21.4/x64/src/bytes/buffer.go:261 +0x106
  github.com/mochi-mqtt/server/v2.(*Client).WritePacket.func1()
      /home/runner/go/pkg/mod/github.com/mochi-mqtt/server/v2@v2.4.3/clients.go:609 +0x2c4
  github.com/mochi-mqtt/server/v2.(*Client).WritePacket()
      /home/runner/go/pkg/mod/github.com/mochi-mqtt/server/v2@v2.4.3/clients.go:616 +0x6b9
  github.com/mochi-mqtt/server/v2.(*Client).WriteLoop()
      /home/runner/go/pkg/mod/github.com/mochi-mqtt/server/v2@v2.4.3/clients.go:198 +0x1c4
  github.com/mochi-mqtt/server/v2.(*Server).NewClient.func1()
      /home/runner/go/pkg/mod/github.com/mochi-mqtt/server/v2@v2.4.3/server.go:237 +0x33

Previous read at 0x00c00137f0e0 by goroutine 1115:
  bytes.(*Buffer).Bytes()
      /opt/hostedtoolcache/go/1.21.4/x64/src/bytes/buffer.go:54 +0x804
  github.com/mochi-mqtt/server/v2.(*Client).WritePacket()
      /home/runner/go/pkg/mod/github.com/mochi-mqtt/server/v2@v2.4.3/clients.go:627 +0x848
  github.com/mochi-mqtt/server/v2.(*Server).processPubrec()
      /home/runner/go/pkg/mod/github.com/mochi-mqtt/server/v2@v2.4.3/server.go:1089 +0x3a4
  github.com/mochi-mqtt/server/v2.(*Server).processPacket()
[...]

PR #347, opened 2023-12-16
Fix data race issue with write buffer

thedevop:
Ensure write buffer is not the one used by Hook. As reported in #346.

### DIFF
--- a/clients.go
+++ b/clients.go
@@ -598,12 +598,7 @@ func (cl *Client) WritePacket(pk packets.Packet) error {
 
 		// there are more writes in the queue
 		if cl.Net.outbuf == nil {
-			if l := buf.Len(); l < cl.ops.options.ClientNetWriteBufferSize {
-				cl.Net.outbuf = buf
-				return int64(l), nil
-			}
-
-			return buf.WriteTo(cl.Net.Conn)
+			cl.Net.outbuf = new(bytes.Buffer)
 		}
 
 		n, _ = buf.WriteTo(cl.Net.outbuf) // will always be successful


x20080406 (approved):

werbenhu:
@thedevop @mochi-co @x20080406 @dgduncan 
In the current processing logic, regardless of network congestion, packets are first written into the write buffer and then into the conn. Even with only one client at the moment, which sends one packet per second, the packet is still buffered before being written into the connection.

I think we need to reconsider this approach. Is it possible to consider using the write buffer only under specific conditions? For example, conditions such as len(cl.State.outbound) > 25% * MAX or buf.len() > 2K.

At the same time, we also need to consider that WritePacket() is called in both the read and write coroutines of each client, which may involve data races.

I would like to hear your opinions.

thedevop:
@werbenhu, if there is no congestion and there were no buffer waiting to be flushed, the packets are directly written to the connection without buffering. Using your example of `Even with only one client at the moment, which sends one packet per second, the packet is still buffered before being written into the connection.`, the buffer will not kick in.

werbenhu (commented):
@thedevop you're right! I missed that.  Let's address this bug first and then optimize the rest later.

werbenhu (approved):

werbenhu:
@thedevop  I think encapsulating two WritePacket functions is a good choice—one to handle packets within the outbound process and another to handle packets written by the write coroutine. We only need to use the write buffer when processing outbound packets, which can help avoid data race issues and allow for buffer reuse. This way, the code can also be cleaner.

thedevop:
@werbenhu, the data race issue was from calling hook after the actual write, where it's not protected by the lock. So having 2 WritePacket functions, we still can not reuse the same buffer for the packet as the buffer for the connection.

Note, during packet encoding, with #343, writing of packet payload is reduced by 1. There maybe more opportunities to optimize.

mochi-co:
Looks good to me, I love a one-line change :)

PR #351, opened 2023-12-25
Implement File based configuration

mochi-co:
Following our discussions and @dgduncan's work in #309, here is a PR which implements a new file-based configuration for the broker. 

A runnable example can be found in `examples/config`.

**How it works**
Rather creating a layer which adds listeners and hooks as they are in the main.go files, I have opted instead to update the hooks and listeners so that they can be applied directly through the server options.

The server options, capabilities, and compatibilities transfer directly from yaml/json into their normal options struct, as though it were being provided as a value created in main.go. Refer to `examples/config/config.yaml` (or `config.json`) for all options.
 
The listeners are defined using a slice Listener.Config, specifying the Type, ID, and Address. TLS can also be specified although I haven't tried it, so it may need a bit of work.

Hooks are defined individually, and split by category. As above, see config.yaml for a full list of options for each hook. This was a bit tricky to implement in a respectable manner, but in the end I decided to simply give hooks their own configuration struct, and allow only one type of each hook, which allows us to provide custom parameters for each (e.g. Auth and Storage.Redis).

I have updated the Dockerfile and added a `cmd/docker.main.go` as per @dgduncan's original work. I have not tested it.

**Caveats**
- If you use file-based configuration, you can only have one of each hook type. However, this should be fine for 99% of users.
- You can only use built in hooks with file-based configuration, as the type needs to be known in order for it to be applied.
- You can only use built in listeners, for the reasons above.

Contains breaking changes for the way listeners are initialized (id and address are now in Config).

If you wish to use custom hooks and listeners, you will currently need to modify the code to support these, or just use the standard system of applying them in a custom go file.

I suspect I've missed some edge-cases too, but thought I'd submit it for review :)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -19,3 +19,4 @@ WORKDIR /
 COPY --from=builder /app/mochi .
 
 ENTRYPOINT [ "/mochi" ]
+CMD ["/cmd/docker", "--config", "config.yaml"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -11,21 +11,11 @@ RUN go mod download
 
 COPY . ./
 
-RUN go build -o /app/mochi ./cmd
-
+RUN go build -o /app/mochi ./cmd/docker
 
 FROM alpine
 
 WORKDIR /
 COPY --from=builder /app/mochi .
-
-# tcp
-EXPOSE 1883
-
-# websockets
-EXPOSE 1882
-
-# dashboard
-EXPOSE 8080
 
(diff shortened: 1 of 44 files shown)


dgduncan:
@mochi-co great job carrying on the work! I'll take a look at this asap.

torntrousers:
Would this file based config enable running a broker as an mqtt bridge to another broker with just a config file? Something like all publishes to this broker get published on to a remote broker under some topic tree named in the config file?

PR #355, opened 2023-12-28
Buffer optimizations

thedevop:
Few buffer optimizations:
1. Avoid creating/writing to outbound buffer if outbound was nil and pkt is larger than ClientNetWriteBufferSize
2. Use mempool for Properties Encode as well
3. Use the more efficient Write instead of WriteTo for Buffer to Buffer writes (as it won't be partially successful)

Benchmark comparing Write vs WriteTo (the performance difference is reverse proportion to the packet size, using 128 bytes for benchmark as vast majority packets without payload should be less than that):
```go
func BenchmarkWrite(b *testing.B) {
	const PacketSize = 128
	d := make([]byte, PacketSize)
	if _, err := rand.Read(d); err != nil {
		panic(err)
	}
	dst := new(bytes.Buffer)

	b.ResetTimer()
	b.ReportAllocs()
	for i := 0; i < b.N; i++ {
		src := bytes.NewBuffer(d)
		dst.Write(src.Bytes())
		dst.Reset()
	}
}

func BenchmarkWriteTo(b *testing.B) {
	const PacketSize = 128
	d := make([]byte, PacketSize)
	if _, err := rand.Read(d); err != nil {
		panic(err)
	}
	dst := new(bytes.Buffer)

	b.ResetTimer()
	b.ReportAllocs()

	for i := 0; i < b.N; i++ {
		src := bytes.NewBuffer(d)
		src.WriteTo(dst)
		dst.Reset()
	}
}
```

```
cpu: Intel(R) Core(TM) i7-7920HQ CPU @ 3.10GHz
BenchmarkWrite-8     	163843377	         7.306 ns/op	       0 B/op	       0 allocs/op
BenchmarkWriteTo-8   	100000000	        12.15 ns/op	       0 B/op	       0 allocs/op
```

### DIFF
--- a/clients.go
+++ b/clients.go
@@ -582,7 +582,7 @@ func (cl *Client) WritePacket(pk packets.Packet) error {
 		return packets.ErrPacketTooLarge // [MQTT-3.1.2-24] [MQTT-3.1.2-25]
 	}
 
-	n, err := func() (n int64, err error) {
+	n, err := func() (int64, error) {
 		cl.Lock()
 		defer cl.Unlock()
 		if len(cl.State.outbound) == 0 {
@@ -591,9 +591,9 @@ func (cl *Client) WritePacket(pk packets.Packet) error {
 			}
 
 			// first write to buffer, then flush buffer
-			n, _ = buf.WriteTo(cl.Net.outbuf) // will always be successful
+			n, _ := cl.Net.outbuf.Write(buf.Bytes()) // will always be successful
 			err = cl.flushOutbuf()
-			return
+			return int64(n), err
 		}
 
 		// there are more writes in the queue
@@ -604,13 +604,13 @@ func (cl *Client) WritePacket(pk packets.Packet) error {
 			cl.Net.outbuf = new(bytes.Buffer)
 		}
 
-		n, _ = buf.WriteTo(cl.Net.outbuf) // will always be successful
+		n, _ := cl.Net.outbuf.Write(buf.Bytes()) // will always be successful
 		if cl.Net.outbuf.Len() < cl.ops.options.ClientNetWriteBufferSize {
-			return
+			return int64(n), nil
 		}
 
 		err = cl.flushOutbuf()
-		return
+		return int64(n), err
 	}()
 	if err != nil {
 		return err
--- a/clients.go
+++ b/clients.go
@@ -598,6 +598,9 @@ func (cl *Client) WritePacket(pk packets.Packet) error {
 
 		// there are more writes in the queue
 		if cl.Net.outbuf == nil {
+			if buf.Len() >= cl.ops.options.ClientNetWriteBufferSize {
+				return buf.WriteTo(cl.Net.Conn)
+			}
 			cl.Net.outbuf = new(bytes.Buffer)
 		}
 
--- a/packets/packets.go
+++ b/packets/packets.go
@@ -348,7 +348,7 @@ func (pk *Packet) ConnectEncode(buf *bytes.Buffer) error {
 
 	pk.FixedHeader.Remaining = nb.Len()
 	pk.FixedHeader.Encode(buf)
-	_, _ = nb.WriteTo(buf)
+	buf.Write(nb.Bytes())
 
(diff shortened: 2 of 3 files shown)


mochi-co:
This looks good to my eyes, but would like to confirm with @werbenhu before we merge

werbenhu:
@thedevop I added two benchmark test cases with different results. This scenario should be closer to our use case, where we are writing from one buffer to another. Take a look at the test results on your env.

```go
func BenchmarkWrite(b *testing.B) {
	const PacketSize = 128
	d := make([]byte, PacketSize)
	if _, err := rand.Read(d); err != nil {
		panic(err)
	}
	dst := new(bytes.Buffer)
	
	b.ResetTimer()
	b.ReportAllocs()
	for i := 0; i < b.N; i++ {
		src := bytes.NewBuffer(d)
		dst.Write(src.Bytes())
		dst.Reset()
	}
}

func BenchmarkWriteTo(b *testing.B) {
	const PacketSize = 128
	d := make([]byte, PacketSize)
	if _, err := rand.Read(d); err != nil {
		panic(err)
	}
	dst := new(bytes.Buffer)

	b.ResetTimer()
	b.ReportAllocs()

	for i := 0; i < b.N; i++ {
		src := bytes.NewBuffer(d)
		src.WriteTo(dst)
		dst.Reset()
	}
}

func BenchmarkWrite2(b *testing.B) {
	const PacketSize = 128
	d := make([]byte, PacketSize)
	if _, err := rand.Read(d); err != nil {
		panic(err)
	}
	dst := new(bytes.Buffer)
	src := bytes.NewBuffer(d)

	b.ResetTimer()
	b.ReportAllocs()
	for i := 0; i < b.N; i++ {
		dst.Write(src.Bytes())
		dst.Reset()
	}
}

func BenchmarkWriteTo2(b *testing.B) {
	const PacketSize = 128
	d := make([]byte, PacketSize)
	if _, err := rand.Read(d); err != nil {
		panic(err)
	}
	dst := new(bytes.Buffer)

	src := bytes.NewBuffer(d)
	b.ResetTimer()
	b.ReportAllocs()

	for i := 0; i < b.N; i++ {
		src.WriteTo(dst)
		dst.Reset()
	}
}
```
```
BenchmarkWrite-4        100000000               10.02 ns/op            0 B/op          0 allocs/op
BenchmarkWriteTo-4      70746446                14.61 ns/op            0 B/op          0 allocs/op
BenchmarkWrite2-4       125998378                8.958 ns/op           0 B/op          0 allocs/op
BenchmarkWriteTo2-4     300416527                3.880 ns/op           0 B/op          0 allocs/op
```

thedevop:
@werbenhu , Buffer once is read, it will no longer be available to read again. Hence
```go
	for i := 0; i < b.N; i++ {
		src.WriteTo(dst)
		dst.Reset()
	}
```
will only write its content to dst the 1st time, subsequent writes within the for loop will be a 0 byte write. That's the reason in the original benchmark the src is reinitialize within the for loop.

Also, if we look at the source for Buffer.WriteTo, it actually calls the dst.Write with some additional checks and error handling. So `dst.Write(src.Bytes())` will always outperform `src.WriteTo(dst)`
```go
func (b *Buffer) WriteTo(w io.Writer) (n int64, err error) {
	b.lastRead = opInvalid
	if nBytes := b.Len(); nBytes > 0 {
		m, e := w.Write(b.buf[b.off:])
		if m > nBytes {
			panic("bytes.Buffer.WriteTo: invalid Write count")
		}
		b.off += m
		n = int64(m)
		if e != nil {
			return n, e
		}
		// all bytes should have been written, by definition of
		// Write method in io.Writer
		if m != nBytes {
			return n, io.ErrShortWrite
		}
	}
	// Buffer is now empty; reset.
	b.Reset()
	return n, nil
}
```

werbenhu (approved):
@thedevop  You are right; it does look good.

mochi-co:
I'm also very happy with this, so nice work @thedevop! Thank you @werbenhu for your review. I'll merge it in!

Issue #356, opened 2023-12-31
OnACLCheck BUG


mrhb6006:
![image](https://github.com/mochi-mqtt/server/assets/45798671/d7474640-78a0-44e0-bc1c-19bdb3f659d7)
 here in the third input at aclCheck function should be "true"  (mqtt package - server.go)
 
 Naturally, this parameter is expected to have the correct value ("true")  during publish
It caused a bug in my program at acl check

thedevop:
@mrhb6006 , you're correct, the write parameter should be set to true. Would you like to create a PR?

mrhb6006:
> @mrhb6006 , you're correct, the write parameter should be set to true. Would you like to create a PR?

tnx 
Done
https://github.com/mochi-mqtt/server/pull/357

mrhb6006:
how long do you think my pull request will take to merge? i need it in my project and my project doesn't work correctly now
it is a bad bug

thedevop:
@mrhb6006 , I have approved the PR. But @mochi-co is needed for final approve and merge. He is very busy lately, so I'm not able to offer an ETA.

In the meantime, you can vendor this module and make modification locally.

mochi-co:
I maybe wrong, but I'm not 100% sure this assumption is correct. 

`publishToClient` is used to write out packets to subscribers - in this case the ACL check should be for the subscribed client to _read_ from the topic they are subscribed to (some ACL configurations allow a user to write to a topic but not read from it, such as when issuing sensitive user information that other nodes shouldn't be able to see).

The ACL check for sending messages to a topic is handled in processPublish:
```go
if !cl.Net.Inline && !s.hooks.OnACLCheck(cl, pk.TopicName, true) {
```

Incidentally there is an additional ACL read check that occurs in processSubscribe to prevent users without read access from subscribing to specific topics. 

@mrhb6006 could you describe the exact problem you are seeing? And perhaps provide an example of your ACL configuration? This way we may be able to narrow down the cause.

thedevop:
@mrhb6006 , my bad. @mochi-co is correct. The ACL check in publishToClient is the subscriber and not the publisher.

mrhb6006:
@mochi-co @thedevop 
![image](https://github.com/mochi-mqtt/server/assets/45798671/ec59aa20-8e7c-4a5d-ba9a-7c6389d0b62e)

[...]

PR #357, opened 2024-01-01
fix(acl) : fix OnACLCheck parameter in publishToClient method

mrhb6006:
fix this issue https://github.com/mochi-mqtt/server/issues/356


thedevop (approved):

thedevop (changes requested):
This check is for subscriber, hence should be write should be false.

PR #359, opened 2024-01-14
Allow configurable max stored qos > 0 messages

thedevop:
Address issue #340.

Issue #200, opened 2023-04-17
Concurrency issues when reading and writing retain messages 


yh742:
Hi there,

I'm trying to embed the broker into an application, but I've noticed while testing the library, the race detector goes off when publishing and subscribing to retain messages at the same time. I believe the issue is that there's no locks here:

https://github.com/mochi-co/mqtt/blob/7bd7bd5087c40f96015a61d01ce7ae4e6951c807/topics.go#L426-L428

And there's a lock here:

https://github.com/mochi-co/mqtt/blob/7bd7bd5087c40f96015a61d01ce7ae4e6951c807/topics.go#L354-L377

mochi-co:
Hi @yh742 - could you paste the race condition logs so I can have a closer look? Thanks!

yh742:
Sure, I have attached the log. For a bit of context, I'm trying to inject a retain message to a topic when a device connects and remove the retain message when disconnection occurs.

```
WARNING: DATA RACE
Write at 0x00c0000251e8 by goroutine 100:
  github.com/mochi-co/mqtt/v2.(*TopicsIndex).RetainMessage()
      /home/sean/go/pkg/mod/github.com/mochi-co/mqtt/v2@v2.2.7/topics.go:372 +0x37d
  github.com/mochi-co/mqtt/v2.(*Server).retainMessage()
      /home/sean/go/pkg/mod/github.com/mochi-co/mqtt/v2@v2.2.7/server.go:758 +0xe4
  github.com/mochi-co/mqtt/v2.(*Server).processPublish()
      /home/sean/go/pkg/mod/github.com/mochi-co/mqtt/v2@v2.2.7/server.go:715 +0x769
  github.com/mochi-co/mqtt/v2.(*Server).processPacket()
      /home/sean/go/pkg/mod/github.com/mochi-co/mqtt/v2@v2.2.7/server.go:569 +0x2a7
  github.com/mochi-co/mqtt/v2.(*Server).InjectPacket()
      /home/sean/go/pkg/mod/github.com/mochi-co/mqtt/v2@v2.2.7/server.go:658 +0xa4
  github.com/mochi-co/mqtt/v2.(*Server).Publish()
      /home/sean/go/pkg/mod/github.com/mochi-co/mqtt/v2@v2.2.7/server.go:641 +0x184
  github.com/mochi-co/mqtt/v2.(*Server).Publish-fm()
      <autogenerated>:1 +0x99
[...]

PR #360, opened 2024-01-14
Replace DefaultServerCapabilities with NewDefaultServerCapabilities()…

thedevop:
… to avoid data race

In tests, Capabilities maybe altered. However, DefaultServerCapabilities is a single global instance, where it maybe embedded in running server instances. So when value is altered, tests may fail with data race issue.

Here is an example:
In TestServerProcessSubscribeACLCheckDenyObscure:
https://github.com/mochi-mqtt/server/blob/65c78534dcb5861dc8ea2163ac58ce603b6e9881/server_test.go#L2790-L2794
and TestServerClearExpiredRetained
https://github.com/mochi-mqtt/server/blob/65c78534dcb5861dc8ea2163ac58ce603b6e9881/server_test.go#L3328
are seemingly different tests, but they cause interferences and resulted in data race issue.

This surfaced in the failed test for PR #359.

Note, this is mostly a test issue, as normally one would (should) not alter server capabilities once server started serving.

This may also be the cause for data race issue seen in #200.

### DIFF
--- a/server.go
+++ b/server.go
@@ -33,21 +33,8 @@ const (
 )
 
 var (
-	// DefaultServerCapabilities defines the default features and capabilities provided by the server.
-	DefaultServerCapabilities = &Capabilities{
-		MaximumSessionExpiryInterval: math.MaxUint32, // maximum number of seconds to keep disconnected sessions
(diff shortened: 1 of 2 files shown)


mochi-co:
This is a very nice fix, thank you @thedevop. I really like what you've done here.

Issue #370, opened 2024-02-19
[badgerdb] vlog growing unbounded - consider adding GC and exposing options


woodzy:
I noticed that BadgerDB storage started growing unbounded (it was many files and several gigabytes quite quickly). Looks like the Badgerhold/Badger Storage example doesn't provide a way for the user to utilize GC correctly per Badgers docs. https://github.com/dgraph-io/badger/blob/main/db.go#L1225

Fairly sure the option tuning is probably useful to expose as well somehow -- for example ValueLogFileSize has a [pretty large default](https://github.com/dgraph-io/badger/blob/6acc8e801739f6702b8d95f462b8d450b9a0455b/options.go#L180)

I didn't plumb up a full fix, just copied the default and started experimenting, but this showed that something like option tuning and GC can resolve the infinite growth modified a copy of the default to include some signalling and a regular GC call.

I would recommend at least adding the GC as a default. I would also suggest lowering some of the default or exposing them.

```
        go func() {
                ticker := time.NewTicker(5 * time.Minute)
                defer ticker.Stop()
                for { 
                        select {
                                case <- ticker.C:
                                        _ = h.db.Badger().RunValueLogGC(0.7)
                                        continue
                                case <- h.doneCh:
                                        return
                        }
                }
        }()
```

werbenhu:
I've checked the official documentation of Badger regarding [garbage collection](https://dgraph.io/docs/badger/get-started/#garbage-collection). @woodzy , you're right. Since neither of us has experience using BadgerDB in a production environment, if you could, it would be great if you could submit a PR. This would be very helpful for most users who use BadgerDB as a persistent database.

werbenhu:
@woodzy Can this #371 fix the issue?

woodzy:
Thanks! These look like better implementations of my local workarounds during experimentation.

mochi-co:
This should be fixed in `v2.6.0` - please let us know!

OpenJarvisAI:
<img width="512" alt="image" src="https://github.com/mochi-mqtt/server/assets/136460643/2b847461-aeec-4cac-ab79-1cc352d30017">

It not fix, still growing on mem

werbenhu:
@woodzy @OpenJarvisAI Could this issue be closed?

werbenhu:
The issue has been resolved, and due to the lack of feedback over an extended period, we are closing this issue. If there are any further problems, please reopen it.

Issue #369, opened 2024-02-13
Hi, what is the simplest way to make messages can be restored when server cut off?


luohao123:
Currently the messages that not sent to client yet will be wiped out when server down.

I previous using badge, but that very terriable, causing about 5GB trash don't know what's file to my server, almost make my server bumm out.

Can just make offline message restore be an default behavior? Really need it!

werbenhu:
@luohao123 Since most of our developers lack experience using BadgerDB as persistent data in a production environment, take a look at the configuration modifications and the addition of garbage collection described in #370 to see if they would be helpful for your use of BadgerDB.

werbenhu:
@luohao123 Could you please check if version v2.6.0 resolves this issue?

luohao123:
It's not work , dude.

werbenhu:
@luohao123 The latest progress is discussed here: #375

werbenhu:
@luohao123 Could this issue be closed?

werbenhu:
The issue has been resolved, and due to the lack of feedback over an extended period, we are closing this issue. If there are any further problems, please reopen it.

Issue #363, opened 2024-01-21
After enabled badger, the vlog file up to 700M one day and 4GB one week


lucasjinreal:
This is too big for my tiny server.

I want keep the lefted message to restore after shutdown server, but this size if far out of my expectations.

Why it is so big? And how to reduce it? (it might about badger but stores what messages left to users who use it, so post the issue here)

mochi-co:
Hi @lucasjinreal - did you check https://dgraph.io/docs/badger/get-started/#memory-usage ? You should be able to apply these configurations in the Badger hook options. Maybe something like this:

```go
// make sure to import the extra badger packages:
import badgerdb "github.com/dgraph-io/badger"
import "github.com/timshannon/badgerhold"

// some complex nesting of different structs called 'Options'
err := server.AddHook(new(badger.Hook), &badger.Options{
  Path: badgerPath,
  Options: &badgerhold.Options{
  	Options: badgerdb.Options{
  		ValueLogFileSize: 256000,
  	},
  },
})
```

This will give access to all the options in https://github.com/dgraph-io/badger/blob/main/options.go

Let me know if it works, if so we can update the example file.

lucasjinreal:
@mochi-co  thanks for reply, should I update mochi lib?

mochi-co:
@lucasjinreal Try updating your main.go with your badger hook and see if it fixes the issue :)

lucasjinreal:
I got no such field in Options.

also, why does vlog file has relation to the memory usage?

mochi-co:
You might be missing an import - see this example:
```go
// SPDX-License-Identifier: MIT
// SPDX-FileCopyrightText: 2022 mochi-mqtt, mochi-co
// SPDX-FileContributor: mochi-co

package main

import (
	"log"
	"os"
	"os/signal"
	"syscall"

	mqtt "github.com/mochi-mqtt/server/v2"
	"github.com/mochi-mqtt/server/v2/hooks/auth"
	"github.com/mochi-mqtt/server/v2/hooks/storage/badger"
	"github.com/mochi-mqtt/server/v2/listeners"

	badgerdb "github.com/dgraph-io/badger"
	"github.com/timshannon/badgerhold"
)

func main() {
	badgerPath := ".badger"
	defer os.RemoveAll(badgerPath) // remove the example badger files at the end

	sigs := make(chan os.Signal, 1)
	done := make(chan bool, 1)
[...]

PR #371, opened 2024-02-27
Add BadgerDB garbage collection.

werbenhu:
For issues #370, #369, and #363. I've tested it myself under the condition of setting the log file size to 1MB. With 1000 clients repeatedly connecting and disconnecting with the same client ID, if GC is not enabled, the log file continues to grow indefinitely. Enabling the GC loop resolves this issue.

### DIFF
--- a/examples/persistence/badger/main.go
+++ b/examples/persistence/badger/main.go
@@ -11,10 +11,12 @@ import (
 	"syscall"
 	"time"
 
+	badgerdb "github.com/dgraph-io/badger"
 	mqtt "github.com/mochi-mqtt/server/v2"
 	"github.com/mochi-mqtt/server/v2/hooks/auth"
 	"github.com/mochi-mqtt/server/v2/hooks/storage/badger"
 	"github.com/mochi-mqtt/server/v2/listeners"
+	"github.com/timshannon/badgerhold"
 )
 
 func main() {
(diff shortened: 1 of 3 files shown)


examples/persistence/badger/main.go
```diff
@@ -31,8 +34,20 @@ func main() {
 	server := mqtt.New(nil)
 	_ = server.AddHook(new(auth.AllowHook), nil)
 
+	// AddHook adds a BadgerDB hook to the server with the specified options.
+	// GcInterval specifies the interval at which BadgerDB garbage collection process runs.
+	// Refer to https://dgraph.io/docs/badger/get-started/#garbage-collection for more information.
 	err := server.AddHook(new(badger.Hook), &badger.Options{
-		Path: badgerPath,
+		Path:       badgerPath,
+		GcInterval: 3 * time.Minute, // Set the interval for garbage collection.
```
dgduncan, line 42: Have you tested this with a shorter `GcInterval` and the impact that has on performance? If the GC functions as a "stop the world" type and with higher client counts it may be a bit too high.

werbenhu: As far as I know, Badger GC would not reclaim the latest value log, which is a read-write log. It only works on the value logs that have become "immutable". So there should be no STW (Stop-The-World) issue.

GC certainly comes with a cost. According to the Badger official documentation: "It is recommended to be called during periods of low activity in your system, or periodically." Regarding the GcInterval, we are just providing an example for reference here. As for GcInterval or other options, users may need to optimize based on their actual scenarios.

dgduncan (approved):

mochi-co:
@werbenhu @dgduncan This generally looks like a great enhancement to me (on my phone at least), and I'm going to try to look at it as soon as I can. Currently I am on the other side of the world (Tasmania) and I've had very little signal or internet for the last couple of weeks. I will be returning to more populated areas in a week. I also hope to finalise some of the other improvements we've discussed like the file based config. Thank you for keeping everything going!

mochi-co (approved):
This all looks very nice to me 👍🏻

OpenJarvisAI:
How to use it?

Can 't ust make it a default config???

PR #379, opened 2024-03-28
Ability to limit maximum number of clients

thedevop:


werbenhu (dismissed):

server.go
```diff
@@ -414,6 +416,16 @@ func (s *Server) attachClient(cl *Client, listener string) error {
 	}
 
 	cl.ParseConnect(listener, pk)
+	if atomic.LoadInt64(&s.Info.ClientsConnected) >= s.Options.Capabilities.MaximumClients {
+		if cl.Properties.ProtocolVersion < 5 {
+			s.SendConnack(cl, packets.ErrServerUnavailable, false, nil)
+		} else {
+			s.SendConnack(cl, packets.ErrQuotaExceeded, false, nil)
```
mochi-co, line 423: Maybe` ErrServerBusy` would be more appropriate, what do you think?

mochi-co (changes requested):

server.go
thedevop: Sure, I made the change to use ErrServerBusy instead.

mochi-co (approved):
@thedevop Sorry I really thought I'd approved this already. Looks great, approving and merging! Thank you

Issue #377, opened 2024-03-26
How to send topics posted by specific users only to specific subscribed users?


leijux:
The location of the hook:

https://github.com/mochi-mqtt/server/blob/47162a3770a38567230d28bbc79477a5c95aceb2/server.go#L1006

Use scenario: 

used to realize the filtering of clients, so that packets are forwarded only to clients that meet the conditions

My current solution:

use OnPacketEncode to filter clients. For clients that do not meet the conditions, I set the FixedHeader.Type field of the packet to packets.WillProperties. However, this approach results in recording error errors.

```go
func (h *Hook) OnPacketEncode(cl *mqtt.Client, pk packets.Packet) packets.Packet {
	switch {
	case pk.TopicName != topicSet && pk.TopicName != topicSetResp:
		return pk
	case pk.Origin == "":
		return pk
	case pk.FixedHeader.Type != packets.Publish:
		return pk
	}
	if pk.TopicName == topicSet && pk.Origin != "mqtt_"+cl.ID { 
		pk.FixedHeader.Type = packets.WillProperties
		return pk
	}
	return pk
}
```

werbenhu:
@leijux If I understand correctly, you want to add some special tags while parsing the packet, so that you can filter them based on these tags during packet processing, right?

If that's the case, I suggest you store your own tag information in cl.Properties.Props.User. Here, you can save some Key-Value data.

leijux:
> @leijux If I understand correctly, you want to add some special tags while parsing the packet, so that you can filter them based on these tags during packet processing, right?
> 
> If that's the case, I suggest you store your own tag information here in `cl.Properties.Props.User`, where you can save some Key-Value data for those users you want to filter.

用户属性可以作为筛选条件，但是我不知道如何在数据包转发阶段对转发行为进行控制。如下图，主题中有三个客户端，我只想将数据发送给其中的某个客户端，我当前的实现是使用 OnPacketEncode hook 函数，将 包的 FixedHeader.Type 设置为无效的类型，让数据包无法转发出去，但是这样会引发 error 错误，有没有更好的办法，感谢。

![image](https://github.com/mochi-mqtt/server/assets/40145943/053ba851-ef90-4d54-83de-503fa9b1b8a3)

werbenhu:
@leijux 你这种的话，实现hook.OnACLCheck就行了

leijux:
[...]

PR #383, opened 2024-04-07
Add OnSendToSubscriber Event Hooks

leijux:
See https://github.com/mochi-mqtt/server/issues/377


werbenhu:
Would merging the onAclCheck method and OnSendToSubscriber into one make it cleaner and tidier? Just need to modify the parameters of onAclCheck. 
Change
 ```go
func (h *HookBase) OnACLCheck(cl *Client, topic string, write bool) bool
```` 
to 
```go
func (h *HookBase) OnACLCheck(cl *Client, pk packets.Packet, write bool) bool
````

werbenhu:
If we directly modify the parameters of `OnACLCheck`, won't it affect compatibility with older versions? @mochi-co @thedevop , what are your thoughts on this?

thedevop:
> If we directly modify the parameters of `OnACLCheck`, won't it affect compatibility with older versions? @mochi-co @thedevop , what are your thoughts on this?

I do like merging this into OnACLCheck, in addition to the benefits you outlined, it is also a bit more performant, as hooks are not "free".

Note, this capability is only usable for embedded use case as current auth hook doesn't provide mechanism to examine packet.

mochi-co:
@werbenhu @thedevop @leijux I'm also in favour of changing the signature of OnACLCheck, as we will otherwise be creating significant duplicated functionality. However, it will be a breaking change to OnACLCheck (and server references and tests), so I recommend opening a new PR which specifically targets this method.

leijux:
@mochi-co @werbenhu @thedevop  Thank you everyone for your opinions. I agree very much with merging into OnACLCheck. I will close this PR and try to merge into OnACLCheck and submit a new PR.

leijux:
@mochi-co @werbenhu @thedevop  Thank you everyone for your opinions. I agree very much with merging into OnACLCheck. I will close this PR and try to merge into OnACLCheck and submit a new PR.

PR #386, opened 2024-04-12
Fix the bug where inline subscribers do not receive messages after all non-inline clients unsubscribe.

werbenhu:
This PR fixes the issue where inline subscribers do not receive messages on the same topic after all non-inline clients unsubscribe.


thedevop (approved):

mochi-co (approved):
Wow, great find and clean fix. Approving :)

PR #389, opened 2024-04-21
Update README example to reflect new listeners API

thedevop:


mochi-co (approved):
Nice fix, thank you @thedevop!

Issue #390, opened 2024-04-21
Merge 2 version of storm


thedevop:
@mochi-co, we seem to be using 2 different versions of storm: https://github.com/mochi-mqtt/server/blob/57997ef0c1460e4ef359c1b8a6b8b10094ead4b7/go.mod#L7-L8

This is because bolt.go is using older version of gob:
https://github.com/mochi-mqtt/server/blob/57997ef0c1460e4ef359c1b8a6b8b10094ead4b7/hooks/storage/bolt/bolt.go#L18

Can this be updated to also use V3?

werbenhu:
@thedevop @mochi-co It seems like the project github.com/asdine/storm is no longer being maintained, and I've seen many complaints about its slowness in its issues. Should we consider bypassing asdine/storm and directly using bbolt?

thedevop:
In general, I tend to favor modules from the source.

mochi-co:
@thedevop I agree, If we can directly use bbolt that would be better - one less dependancy.

thedevop:
Resolved in #392. Thanks @werbenhu!

PR #392, opened 2024-04-23
Bypassing asdine/storm and directly using bbolt.

werbenhu:
For #390, Bypassing asdine/storm and directly using bbolt.


mochi-co (approved):
This is a huge improvement over the existing implementation, and much easier to read and understand. Thank you so much @werbenhu! I'm going to merge this now.

PR #396, opened 2024-04-28
Add logging level file config

dgduncan:
Adds the ability to configure logging level from the file based config. It does this by creating a logger with a `Levelvar` during file based configuration and parsing the included. If no `logging` options are configured, `INFO` is assumed.


mochi-co (approved):
Love this @dgduncan - I wanted to add it to the initial config release, but never came up with a clear way to implement it. This looks like it will be great. Thank you!

PR #399, opened 2024-05-10
Don't allow inheriting session if username doesn't match

### LINKED ISSUE #398: Don't allow inheriting session unless username matches
The method Server.inheritClientSession will use a persistent session with a matching ID **without comparing the usernames first**. This allows a client (accidentally or maliciously) to take over a session belonging to a different user, if it can guess the session ID.

Most seriously, this means the client would inherit the other user's topic subscriptions. It looks as though this bypasses authorization hooks, so the user could have access to topics it shouldn't; that could be a serious security vulnerability. 

At a minimum, it allows a client to disconnect another client or wipe out its persistent session if it can guess the ID; an annoying DoS.

This seems to be covered by [section 5.4.2 of the MQTT spec](https://docs.oasis-open.org/mqtt/mqtt/v5.0/os/mqtt-v5.0-os.html#_Toc3901267):

>...the implementation should check that the Client is authorized to use the Client Identifier as this gives access to the MQTT Session State (described in [section 4.1]. This authorization check is to protect against the case where one Client, accidentally or maliciously, provides a Client Identifier that is already being used by some other Client.

The fix is pretty simple. I'll post a PR shortly.

snej:
MQTT 5.4.2: ".. the implementation should check that the Client is authorized to use the Client Identifier ... to protect against the case where one Client, accidentally or maliciously, provides a Client Identifier that is already being used by some other Client."

Fixes #398

### DIFF
--- a/server.go
+++ b/server.go
@@ -12,6 +12,7 @@ import (
 	"net"
 	"os"
 	"runtime"
+	"slices"
 	"sort"
 	"strconv"
 	"strings"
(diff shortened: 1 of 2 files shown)


snej:
Ironically the tests failed because they exhibit this bug -- they were using inconsistent usernames between the first and second session. Making the usernames the same fixed the problem.

PR #400, opened 2024-05-10
Added client.StopTime(), to expose `disconnected`

snej:
My persistence hook needs to know what time a client disconnected. The field `Client.State.disconnected` has that information but it's not exposed in the API, so I added a simple public accessor method `Client.StopTime()`.

### DIFF
--- a/clients.go
+++ b/clients.go
@@ -417,6 +417,11 @@ func (cl *Client) StopCause() error {
 	return cl.State.stopCause.Load().(error)
 }
 
+// StopTime returns the the time the client disconnected in unix time, else zero.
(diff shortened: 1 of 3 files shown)


werbenhu:
@snej Perhaps it would be better to add a test case to Client.StopTime().

snej:
Added a check to `TestClientStop()`. I also noticed a race condition in that test (comparing a time against
time.Now) and fixed it to allow a one-second discrepancy.

werbenhu (approved):

mochi-co (approved):
Nice addition! Thank you!

PR #401, opened 2024-05-10
Added `omitempty` to storage structs

snej:
Adding `omitempty` to the JSON metadata of the `storage` struct fields significantly cuts down the size of the marshaled JSON when a hook is persisting state to a database.

I've left it out on fields that, in my experience, never have the default value; but if you feel it should be consistent on all fields, I can add it to the rest.


snej:
My bad, I forgot to run tests and it appears some test expects the missing fields to exist.

werbenhu (approved):
This looks very nice.

mochi-co:
This looks good to me, thank you @snej! Apologies for the long wait.

PR #411, opened 2024-06-19
add description for MaximumClientWritesPending

CendealChen:


CendealChen:
pls see ref [MaximumClientWritesPending建议添加文档说明其意义及影响 #410](https://github.com/orgs/mochi-mqtt/discussions/410)

thedevop:
You may also want to point it out that setting this value very too low may result in lost messages for subscribers, especially for busy topics. One can monitor the metrics MessagesDropped when that occurred.

I do agree the current default is biased towards not dropping messages, perhaps 128 may be sufficient, which translates to a fixed 1KB per connection.

mochi-co:
Always happy to have more extensive documentation, thank you @CendealChen

Issue #416, opened 2024-07-18
Client supplied ID is used to look up existing clients in inheritSession, while the store of existing actually uses the server assigned if one exists.


SimonMacIntyre:
I am still debugging and investigating this, so hopefully it is not human error. Figured I would post in case anyone else is experiencing it or able to easily reproduce it. 

Happens to me on versions `> 2.6.0`, I did not test lower than that because my code is not compatible without some changes, so I decided to try to investigate the issue (or my misconfiguration). 

I am using protocol version 5 clients (mqttjs, specifically). 

Here is my server config:
```go
server := mqtt.New(&mqtt.Options{
		InlineClient: true,
		Capabilities: &mqtt.Capabilities{
			MaximumClients:               math.MaxUint16,
			MaximumSessionExpiryInterval: 0,              // maximum number of seconds to keep disconnected sessions
			MaximumMessageExpiryInterval: 60 * 60 * 24,   // maximum message expiry if message expiry is 0 or over
			ReceiveMaximum:               1024,           // maximum number of concurrent qos messages per client
			MaximumQos:                   1,              // maxmimum qos value available to clients
			RetainAvailable:              1,              // retain messages is available
			MaximumPacketSize:            0,              // no maximum packet size
			TopicAliasMaximum:            math.MaxUint16, // maximum topic alias value
			WildcardSubAvailable:         1,              // wildcard subscriptions are available
			SubIDAvailable:               1,              // subscription identifiers are available
			SharedSubAvailable:           1,              // shared subscriptions are available
			MinimumProtocolVersion:       3,              // minimum supported mqtt version (3.0.0)
			MaximumClientWritesPending:   1024 * 8,
		},
	})
```

Here is my client settings:
```javascript
 keepalive: 60,
  reconnectPeriod: 5000,
  clean: false, //I've tried true as well, and with no assigned client id below
  clientId: 'bob',
  debug: true,
  rejectUnauthorized: false,
```

Here is a screenshot of both my client simultaneously connected, and not being forcefully disconnected (also no client takeover occurred according to my logs. 
<img width="1039" alt="image" src="https://github.com/user-attachments/assets/9e0750cf-013b-4ec1-9306-a78a571de860">

Now if I publish a message, only one of them gets it (the latest to connect I think). Now this does seem like a client takeover. 

So the 2 issues are:
- Client who was taken over, is not disconnected
- client event of this exchange is not set or happening (I log client takeovers, and not being detected)

[...]

PR #417, opened 2024-07-19
use cl.ID instead of pk.Connect.ClientIdentifier when looking for existing clients in inheritClientSession

SimonMacIntyre:
Fix https://github.com/mochi-mqtt/server/issues/416

The reason I want to open this PR:
* cl.ID is set from the `pk.Connect.ClientIdentifier` in [clients.go](https://github.com/mochi-mqtt/server/blob/82c96fa4e3f20259c2d947bb87eb8dc885ef6824/clients.go#L236-L236), It also is the same value as the assignedClientId if there is one. Therefore it is more correct, but also for the reasons below:
* This allows my custom implementation to rely on `cl.ID` and override it and the assigned id returned to the client(I assign intentional, known, client ids to specific clients. We want to control the client ids based on the authenticating client.) 

All tests still pass, including the existing client takeover tests.

@mochi-co  Do you think the existing tests are good for this, or do we want some kind of regression test case? I was looking at the tests to add one but to be honest, it is a bit beyond my Go skill level at the moment.

### DIFF
--- a/server.go
+++ b/server.go
@@ -560,7 +560,7 @@ func (s *Server) validateConnect(cl *Client, pk packets.Packet) packets.Code {
 // connection ID. If clean is true, the state of any previously existing client
 // session is abandoned.
 func (s *Server) inheritClientSession(pk packets.Packet, cl *Client) bool {
-	if existing, ok := s.Clients.Get(pk.Connect.ClientIdentifier); ok {
+	if existing, ok := s.Clients.Get(cl.ID); ok {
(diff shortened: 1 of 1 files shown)


SimonMacIntyre:
Huh, not sure I understand the pipeline failure, the tests all worked on this locally 🤔 Anything I can do?

mochi-co:
@SimonMacIntyre Looks like it was just flakey tests 👍🏻

thedevop (approved):

mochi-co:
Nice improvement here, thank you @SimonMacIntyre :)

Issue #453, opened 2025-02-12
The maximum number of connections is significantly lower than other brokers.


xvjialing:
emqx能到19万,mochi-mqtt好像到8万左右就直接程序奔溃了

In a standalone setup on the same machine, the maximum number of connections is significantly lower than other brokers.

EMQX can handle up to 190,000 connections, while Mochi-MQTT crashes at around 80,000 connections.

thedevop:
@xvjialing , can you check what caused the crash? I suspect it's due to Out Of Memory. I run brokers that handle hundreds of thousands clients without issue, but I have to adjust the server settings appropriate for the instance I use.

There are few settings for the broker that can significantly impacts the memory requirement with large number of connections:
1. ClientNetReadBufferSize
2. ClientNetWriteBufferSize
3. Capabilities.MaximumClientWritesPending (each connection will use 8 bytes * the value set for this), for large number of connections, lower this value significantly from the default. I use 128.
4. Also if you have long MaximumMessageExpiryInterval and MaximumSessionExpiryInterval, and often publish QOS > 0 messages to subscribers offline, then this can build up large memory usage as well.

Note, there are some here that can read/write Chinese, but many may not. You may receive quicker response if you post your question in English.

PR #461, opened 2025-03-10
Change MaximumClientWritesPending to 64

lneoe:
Adjust MaximumClientWritesPending default value: from 1024*8 to 64

- Current default value is too high, causing significant memory overhead when maintaining a large number of connections (tens of thousands)
- Testing shows that reducing this value significantly decreases memory usage, optimizing resource utilization

## Technical Impact
- Minimal impact on small-scale deployments
- Significantly improves memory usage in high-concurrency scenarios
- No effect on core functionality and performance

### DIFF
--- a/server.go
+++ b/server.go
@@ -68,7 +68,7 @@ func NewDefaultServerCapabilities() *Capabilities {
 	return &Capabilities{
 		MaximumClients:               math.MaxInt64,  // maximum number of connected clients
 		MaximumMessageExpiryInterval: 60 * 60 * 24,   // maximum message expiry if message expiry is 0 or over
-		MaximumClientWritesPending:   1024 * 8,       // maximum number of pending message writes for a client
+		MaximumClientWritesPending:   64,             // maximum number of pending message writes for a client
 		MaximumSessionExpiryInterval: math.MaxUint32, // maximum number of seconds to keep disconnected sessions
 		MaximumPacketSize:            0,              // no maximum packet size
 		maximumPacketID:              math.MaxUint16,


thedevop:
@lneoe, for my use case (also large number of connections), I also set this to a much smaller value (128) than the default. However, many do not necessary use in environments with large number of connections. For those use cases, it is preferable to have a higher value to prevent possible loss of delivery.

lneoe:
> @lneoe, for my use case (also large number of connections), I also set this to a much smaller value (128) than the default. However, many do not necessary use in environments with large number of connections. For those use cases, it is preferable to have a higher value to prevent possible loss of delivery.

Yes, larger values can prevent message loss, but this consideration is primarily necessary in scenarios with very frequent message transmission and relatively slow consumption rates.
In my actual application scenario, the server maintains approximately 33,000 simultaneous connections, consuming about 5GB of memory that never gets garbage collected. Memory consumption became a bottleneck, preventing the server from creating additional connections. After reducing this value, we now support 50,000 simultaneous connections while consuming only about 2.8GB of memory. Of course, 64 isn't appropriate for every scenario, but the original default value was indeed too large and easily led to excessive memory growth.

thedevop:
Are you using the broker in embedded mode? if so, you can consult #453 for additional server settings that can reduce memory usage.

lneoe:
> Are you using the broker in embedded mode? if so, you can consult #453 for additional server settings that can reduce memory usage.

Yes, I'm using embedded mode, and I've also configured it through setting my own mqtt.Options. The main reason for submitting this PR is considering that such a large default value isn't necessary for most scenarios, and we often use default settings because they typically represent a more universal solution.

Issue #423, opened 2024-07-30
How to configure WSS


ascarl2010:
How to configure WSS,How to configure WSS,How to configure WSS,How to configure WSS,How to configure WSS,How to configure WSS

ascarl2010:
WebSocket Secure

werbenhu:
This issue is the same as #419. @mochi-co @dgduncan @thedevop should we consider implementing this feature in the future?

dgduncan:
@werbenhu Considering we have seen multiple people request it I think it would be a good thing to put higher up on the list of things to do.

PR #472, opened 2025-05-06
Add TLS cert configuration options for listeners on docker configs

kevinta893:
Added some changes to enable secure websocket, mqtt, and http by adding certificate configs in the options files for Docker. Both JSON and YAML are enabled.

Sorry, I am not very familiar with GO so pardon my coding style for it.

I followed the `examples/tls/main.go` file and generalized it for all listener types. You can even specify different certificates and keys per listener endpoint. `tls_cert` is an optional config so if someone wanted both mqtt and mqtts or ws and wss, they can do so on separate ports.

I have a few things I'm not so sure about and would like some feedback on:

1. I have tested the config with TCP (mqtts), Websockets (wss), the Health check (https), and the SysInfo (https), but does it make sense to load the certificate for Unix Sockets and the Mock Listener? (haven't worked with Unix Sockets)
2. Objections to the name `tls_cert`? I figure it would be best to separate them from the existing field `tlsConfig`. I originally had it as simply `ssl` but not sure if the distinction would be confusing.
3. I also have not the experience with CA certs, so for now I think I'll leave it as is and let someone else contribute to it. Though I can see it reasonably extended in the `tls_config`

Thank you

Related Issues:
https://github.com/orgs/mochi-mqtt/discussions/419
https://github.com/mochi-mqtt/server/issues/423

### DIFF
--- a/cmd/docker/main.go
+++ b/cmd/docker/main.go
@@ -1,10 +1,11 @@
 // SPDX-License-Identifier: MIT
 // SPDX-FileCopyrightText: 2023 mochi-mqtt
-// SPDX-FileContributor: dgduncan, mochi-co
+// SPDX-FileContributor: dgduncan, mochi-co, kevinta893
 
 package main
 
 import (
+	"crypto/tls"
 	"flag"
 	"github.com/mochi-mqtt/server/v2/config"
 	"log"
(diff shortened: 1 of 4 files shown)


kevinta893:
Sorry let me close this PR since I wanted to use a different branch for this PR. https://github.com/mochi-mqtt/server/pull/473

PR #485, opened 2025-11-27
Mqtt311 loosen validation

andrewbogo:


dgduncan:
@andrewbogo Out of curiosity what is the reasoning behind loosening validation for v3?

werbenhu:
I do not think we should merge this as-is. Per the MQTT spec, Protocol Name and Protocol Level are a defined pair: "MQIsdp"+3 for MQTT 3.1 and "MQTT"+4 for MQTT 3.1.1. This change removes that check and would allow invalid combinations like "MQTT"+3 or "MQIsdp"+4 to pass.

PR #487, opened 2025-12-12
mqtt over quic support

snac21:


IshanDaga:
@snac21 Happy to consider merging this, since there is a growing interest around MQTT over QUIC in general.
If you could also flesh out the PR message, and outline design / integration decisions, we can move forward with a review.

Issue #488, opened 2025-12-24
Deadlock in `clients.go` due to recursive read-locking


roosenj:
We use this server to test our MQTT code, and we've been seeing some flakey tests related to this bug. The problem occurs in the `GetByListener` method in `clients.go`:

```go
// Len returns the length of the clients map.
func (cl *Clients) Len() int {
	cl.RLock()
	defer cl.RUnlock()
	val := len(cl.internal)
	return val
}

// GetByListener returns clients matching a listener id.
func (cl *Clients) GetByListener(id string) []*Client {
	cl.RLock()
	defer cl.RUnlock()
	clients := make([]*Client, 0, cl.Len())
	for _, client := range cl.internal {
		if client.Net.Listener == id && !client.Closed() {
			clients = append(clients, client)
		}
	}
	return clients
}
```

The deadlock can occur if the following steps occur in order:

1. `GetByListener` gets called and acquires a read lock on `*Clients`
2. Another Goroutine requests a write lock on `*Clients`
3. `cl.Len()` is called from `GetByListener`, which then also attempts to acquire a read lock

The process is now in a deadlock due to the following behavior, as described in the `sync` docs:

> If any goroutine calls [RWMutex.Lock](https://pkg.go.dev/sync#RWMutex.Lock) while the lock is already held by one or more readers, concurrent calls to [RWMutex.RLock](https://pkg.go.dev/sync#RWMutex.RLock) will block until the writer has acquired (and released) the lock, to ensure that the lock eventually becomes available to the writer.

The write lock request blocks `cl.Len()` from acquiring the lock, while the read lock acquired by `GetByListener` cannot be released due to it being blocked by `cl.Len()`'s request for a read lock.

The simple fix for this would be to not use `Len` here, and instead use `len(cl.internal)` directly.

roosenj:
I've created a PR that fixes this specific deadlock. It's only a one line change.

quarksus:
Independently ran into this while stress-testing an embedding project's mTLS work — confirmed it's a real, reproducible deadlock with a minimal, consumer-code-free harness (a real `mqtt.Server` + a real TCP listener, racing repeated connect/disconnect churn against `Close()`, no application code involved): triggers in 33–67 iterations on v2.7.9, zero recurrence in 3000+ iterations after the fix.

[...]

PR #489, opened 2026-01-05
Fix potential deadlock in `GetByListener` function

roosenj:
The deadlock occured when a write lock was requested between the two recursive read locks. The second read lock request blocked, preventing the first read lock from releasing and allowing the write lock, thus creating a deadlock.

This fixes issue #488


werbenhu (dismissed):
This is indeed an issue. Thanks, @roosenj

IshanDaga (dismissed):

roosenj:
seems I pushed our gitlab CI configuration by accident, so I've removed it

PR #493, opened 2026-02-10
fix log spam when EOF error on connection close

martin31821:
currently, the TCP listener logs every connection end, even if it's clean with io.EOF.


werbenhu:
This makes sense, but I am not sure checking only io.EOF is enough, since normal connection close may show up as other errors too. Maybe debug is a better level for expected close-related errors, and if we keep this check, it would be better to use errors.Is(err, io.EOF).

PR #509, opened 2026-08-09
listeners: fix data race between CloseAll and connections being accepted

ifnesi:
### The bug

`attachClient` registers each client with `Listeners.ClientsWg`, and `CloseAll` waits on that group:

```go
// server.go, attachClient
defer s.Listeners.ClientsWg.Done()
s.Listeners.ClientsWg.Add(1)
```

```go
// listeners/listeners.go, CloseAll
l.ClientsWg.Wait()
```

`sync.WaitGroup` documents that a positive `Add` must not run concurrently with `Wait`, and that is exactly what happens when a connection is accepted while the server is closing: the `Add` races the wait that is meant to cover it.

Two consequences:

- Under `-race`, the detector reports a write in `Server.Close` against a read in `Server.EstablishConnection`.
- Without it, `CloseAll` can return while a client is still being attached, so a graceful shutdown does not actually wait for everyone it is supposed to.

The `Done()` was also deferred on the line above the `Add`, which is the usual idiom inverted.

### The fix

Registration moves into `Listeners`, which wraps the establish function in `Serve` — every connection already reaches `attachClient` through there, so this covers all of them in one place.

Shutdown latches under a write lock; clients register under a read lock. No `Add` can follow `Wait`. A connection that arrives after the latch is closed rather than accounted for, which seemed better than logging on every connection that turns up while a server is going down.

Nothing new is exported, and `attachClient` loses two lines.

### Testing

`TestCloseWhileConnecting` opens connections continuously while closing the server, forty times over. It reports a data race against `main` and passes with this change:

```
$ go test -race -run TestCloseWhileConnecting .        # before
WARNING: DATA RACE
--- FAIL: TestCloseWhileConnecting

$ go test -race -run TestCloseWhileConnecting .        # after
ok      github.com/mochi-mqtt/server/v2
```

The existing suite passes under `-race`. Found with the race detector in a downstream test suite, where it failed roughly one run in three and always blamed whichever test happened to be running.


mochi-co (approved):

tomfitzhenry:
> sync.WaitGroup documents that a positive Add must not run concurrently with Wait, and that is exactly what happens when a connection is accepted while the server is closing: the Add races the wait that is meant to cover it.

This isn't quite right.

https://pkg.go.dev/sync#WaitGroup.Add documents that "Note that calls with a positive delta that occur when the counter is zero must happen before a Wait. Calls with a negative delta, or calls with a positive delta that start when the counter is greater than zero, may happen at any time. **Typically this means the calls to Add should execute before the statement creating the goroutine or other event to be waited for.**".

The problem with the existing code is that the goroutine itself calls Add, and so it might not happen before the Wait call.

So the fix would be to move the Add call just before the goroutine spawning.

Or better yet, use  https://pkg.go.dev/sync#WaitGroup.Go as the `Add` docs mention "Callers should prefer WaitGroup.Go.".

This avoids needing the track function, and avoids the need for the new mutex and closed Boolean.

PR #511, opened 2026-08-10
fix: answer a rejected QoS 2 publish with PUBREC, not PUBACK

### LINKED ISSUE #510: QoS 2 publish rejected via OnPublish gets a PUBACK, not a PUBREC (spec violation)
## Bug

When an `OnPublish` hook returns a `packets.Code` error for an MQTT5
publish, `Server.processPublish` always answers with a **PUBACK**,
even when the original publish was **QoS 2**, which must get a
**PUBREC** per the spec.

https://github.com/mochi-mqtt/server/blob/v2.7.9/server.go#L920-L926

```go
} else if cl.Properties.ProtocolVersion == 5 && pk.FixedHeader.Qos > 0 && errors.As(err, new(packets.Code)) {
    err = cl.WritePacket(s.buildAck(pk.PacketID, packets.Puback, 0, pk.Properties, err.(packets.Code)))
```

`packets.Puback` is hardcoded regardless of QoS. The ACL-deny branch
earlier in the same function already selects PUBACK vs PUBREC by QoS
correctly — this is the same logic, just missing here.

## Why it matters

PUBREC is the defined response to a QoS 2 PUBLISH, and 0x97 (Quota
Exceeded) is an explicitly valid PUBREC reason code. A PUBREC with a
reason code >= 0x80 ends that publish negatively — the sender must not
send PUBREL for it. Answering with a PUBACK instead is a real
wire-protocol violation: a strictly compliant sender's QoS 2 state
machine may not know what to do with a PUBACK for a packet ID it never
expects one for.

## Minimal reproduction

Self-contained, no dependencies beyond this module:

```go
package main

import (
	"bufio"
	"bytes"
	"fmt"
	"io"
	"log/slog"
	"net"
	"time"

	mqtt "github.com/mochi-mqtt/server/v2"
	"github.com/mochi-mqtt/server/v2/hooks/auth"
	"github.com/mochi-mqtt/server/v2/listeners"
	"github.com/mochi-mqtt/server/v2/packets"
)

type rejectHook struct {
	mqtt.HookBase
}

func (*rejectHook) ID() string           { return "reject-hook" }
func (*rejectHook) Provides(b byte) bool { return b == mqtt.OnPublish }
func (*rejectHook) OnPublish(cl *mqtt.Client, pk packets.Packet) (packets.Packet, error) {
	return pk, packets.ErrQuotaExceeded // any packets.Code reproduces this
}

func main() {
	server := mqtt.New(&mqtt.Options{Logger: slog.New(slog.NewTextHandler(io.Discard, nil))})
	_ = server.AddHook(&rejectHook{}, nil)

	tcp := listeners.NewTCP(listeners.Config{ID: "t", Address: ":1892"})
	if err := server.AddListener(tcp); err != nil {
		panic(err)
	}
	go func() { _ = server.Serve() }()
	time.Sleep(100 * time.Millisecond)

	conn, err := net.Dial("tcp", "localhost:1892")
	if err != nil {
		panic(err)
	}

	connectPk := packets.Packet{
		FixedHeader:     packets.FixedHeader{Type: packets.Connect},
		ProtocolVersion: 5,
		Connect: packets.ConnectParams{
			ProtocolName:     []byte("MQTT"),
			ClientIdentifier: "repro",
			Clean:            true,
			Keepalive:        60,
		},
	}
	var buf bytes.Buffer
	_ = connectPk.ConnectEncode(&buf)
	conn.Write(buf.Bytes())

	r := bufio.NewReader(conn)
	discardOnePacket(r) // CONNACK

	pubPk := packets.Packet{
		FixedHeader:     packets.FixedHeader{Type: packets.Publish, Qos: 2},
		ProtocolVersion: 5,
		PacketID:        1,
		TopicName:       "t",
		Payload:         []byte("x"),
	}
	var pbuf bytes.Buffer
	_ = pubPk.PublishEncode(&pbuf)
	conn.Write(pbuf.Bytes())

	hb, _ := r.ReadByte()
	packetType := hb >> 4
	remaining, _, _ := packets.DecodeLength(r)
	body := make([]byte, remaining)
	io.ReadFull(r, body)

	fmt.Printf("expected packet type 5 (PUBREC), got packet type %d\n", packetType)
	if packetType == packets.Puback {
		fmt.Println("BUG CONFIRMED: server answered a QoS2 publish's OnPublish-hook rejection with a PUBACK, not a PUBREC")
	}
}

func discardOnePacket(r *bufio.Reader) {
	r.ReadByte()
	remaining, _, _ := packets.DecodeLength(r)
	io.CopyN(io.Discard, r, int64(remaining))
}
```

Output against v2.7.9:
```
expected packet type 5 (PUBREC), got packet type 4
BUG CONFIRMED: server answered a QoS2 publish's OnPublish-hook rejection with a PUBACK, not a PUBREC
```

## Fix

```go
ackType := packets.Puback
if pk.FixedHeader.Qos == 2 {
    ackType = packets.Pubrec
}
err = cl.WritePacket(s.buildAck(pk.PacketID, ackType, 0, pk.Properties, err.(packets.Code)))
```

PR incoming with this exact change plus a regression test.

fcraviolatti:
Fixes #510.

## Bug

`Server.processPublish`'s `OnPublish`-error branch always built a
PUBACK regardless of QoS, even for a QoS 2 publish, which must get a
PUBREC per the spec (PUBREC is the defined response to a QoS 2
PUBLISH; 0x97 Quota Exceeded is an explicitly valid PUBREC reason
code). The ACL-deny branch a few lines above already selects
PUBACK/PUBREC by QoS correctly — this brings the `OnPublish`-error
branch in line with it.

## Fix

```go
ackType := packets.Puback
if pk.FixedHeader.Qos == 2 {
    ackType = packets.Pubrec
}
err = cl.WritePacket(s.buildAck(pk.PacketID, ackType, 0, pk.Properties, err.(packets.Code)))
```

## Test

Added `TestServerProcessPublishOnPublishAckErrorQoS2GetsPubrec`,
mirroring the existing `TestServerProcessPublishOnPublishAckErrorContinue`
but for a QoS 2 publish, asserting the raw PUBREC bytes byte-for-byte.
Confirmed it fails against the pre-fix code and passes against the fix
(mutation-tested locally).


fcraviolatti:
Note: this partially overlaps #476, which also addresses the QoS 2 PUBACK/PUBREC selection. This PR is intentionally narrower: it only fixes the existing `packets.Code` error path so the ack packet type follows the publish's QoS, without changing the unrelated `ErrRejectPacket` handling or MQTT 3.x behavior (#476 touches both of those too). The narrow scope is deliberate, so this specific protocol-correctness issue can be reviewed independently; the PR includes a regression test for the exact case.

PR #514, opened 2026-08-12
Do not refuse SUBSCRIBE/UNSUBSCRIBE when the packet identifier is the server's

ifnesi:
# Do not refuse SUBSCRIBE/UNSUBSCRIBE when the packet identifier is the server's

> **Note on the second commit.** The first commit removed both checks
> outright and argued they could only ever be a false positive. That was
> wrong — there is a true positive, described below — so the second commit
> narrows the fix and restores both existing tests. The history is left in
> place rather than rewritten, since the correction is worth reading.

## What happens

A client's `SUBSCRIBE` is answered with `0x91` (Packet Identifier in use)
whenever the server happens to have an unacknowledged QoS 1 or 2 message
outbound to that client carrying the same identifier.

The commonest way to reach it is the commonest thing in MQTT: a persistent
session with a queued message, and a client that subscribes from its
on-connect callback. Both sides' identifier counters start at 1 on a fresh
connection, so the server's first queued delivery and the client's first
`SUBSCRIBE` collide.

Reproduction, with a stock server, one durable subscriber, and one message
published while it is away:

```
device: CONNACK session_present=true
paho reports: failed to subscribe to topic:
SUBACK reason[0] = 0x91
```

`mosquitto_sub` treats a refused subscription as fatal and exits. Eclipse
Paho (Go) and paho-mqtt 2.1.0 (Python) both take the refusal, stay
connected and go on receiving — so the visible damage depends on the
client, but the refusal itself is wrong in every case.

## Why it is wrong

Packet Identifiers assigned by the Client and by the Server are
independent. MQTT 3.1.1 §2.3.1 and MQTT 5.0 §2.2.1 both say so, and both
give the same example: a Client can send a packet using identifier
`0x1234` while receiving a *different* packet carrying `0x1234` from its
Server.

`cl.State.Inflight` holds entries from **both** spaces:

| entry type | whose identifier | matching a SUBSCRIBE against it |
|---|---|---|
| `Publish` | the server's outbound message | wrong — a different space |
| `Pubrec` | the client's own QoS 2 publish, mid-`PUBREL` exchange | correct — genuinely in use |

`processSubscribe` and `processUnsubscribe` did not distinguish them, so
they refused a legal packet whenever the first row matched.

## The change

Both checks now ask whose identifier they found, which is the same question
`processPublish` already asks a few hundred lines above when it decides
whether an inbound QoS 2 publish is a duplicate:

```go
-	if _, ok := cl.State.Inflight.Get(pk.PacketID); ok {
+	if pki, ok := cl.State.Inflight.Get(pk.PacketID); ok && pki.FixedHeader.Type == packets.Pubrec {
 		code = packets.ErrPacketIdentifierInUse
 	}
```

Two one-line changes. A client that reuses an identifier while its own QoS
2 publish is still in flight is still refused, which is correct and which
the outright removal in the first commit had lost.

## Tests

`TestServerProcessPacketSubscribePacketIDInUse` and
`TestServerProcessPacketUnsubscribePackedIDInUse` are unchanged except for
one field each: the in-flight entry they plant becomes `Type:
packets.Pubrec` rather than `Type: packets.Publish`, which turns each from
the case the specification permits into the case it forbids. Both still
assert `0x91`.

`TestServerProcessSubscribeWithServerInflightPacketID` and
`TestServerProcessUnsubscribeWithServerInflightPacketID` are the other half
of the pair: a server-outbound `Publish` under the same identifier, and the
`SUBACK`/`UNSUBACK` must grant. They fail against `main` with `0x91` where
`0x00` belongs.

The full suite passes.

## One thing not in this change

`processPublish` reads the same map at the top of its QoS 2 handling. Its
`Pubrec` guard means a colliding server-outbound `Publish` entry falls
through to the `Delete` below it, which would drop the server's own pending
delivery. I have not written a test for it and have not confirmed it, so it
is left out rather than guessed at — but it looks like the same root cause
and may be worth a maintainer's eye.


ifnesi:
Correcting myself before anyone spends review time on the original reasoning.

The first commit removed both checks outright and argued they could only ever be a false positive. That is wrong. `cl.State.Inflight` holds a `Pubrec` entry for a client's own in-flight QoS 2 publish, keyed by the identifier the *client* chose — so a client that reuses that identifier for a SUBSCRIBE really does have it in use, and `0x91` is correct there. Deleting the check removed that case along with the bug.

The second commit narrows it instead: the check now asks whether the entry it found is a `Pubrec` (the client's) or a `Publish` (the server's outbound message), which is the same question `processPublish` already asks. Both existing tests come back with one field changed — `Type: packets.Publish` becomes `Type: packets.Pubrec`, turning each into the case that really is a reuse — and the two new tests cover the other side.

Net production change is now two lines, and no test is deleted. I have left the first commit in place rather than force-pushing over it, since the correction seemed worth reading.

mochi-co (approved):

Issue #515, opened 2026-08-12
Is the project still maintained?


ifnesi:
First: thank you for this library. I am building an MQTT 5 broker on top of it and the hook model is the reason it was possible, the design is fantastic, and this is not a complaint about anyone's time.

And I should say plainly who is asking, because it makes me an odd person to be asking it:

What I am building is itself an MQTT broker. It is an application of this one rather than a reimplementation, every packet on the wire is yours, and everything mine adds sits in hooks, but I am a downstream user with an interest, not a neutral observer, and you should read the rest with that in mind. I would rather say so than have you work it out.

I am asking because I need to plan around it, and I would rather ask than assume.

What I can see from the outside:

- The last commit on `main` is 2025-03-01, and the last release (`v2.7.9`) is the same day.
- The last merged pull request is #460, also 2025-03-01.
- There are 22 open pull requests. The oldest is #402, from May 2024.
- Contributors are still opening pull requests and answering each other in the threads, so people are clearly still using it.

That reads like a project whose maintainer's circumstances have changed, which happens to all of us and is nobody's fault. But it leaves everyone downstream guessing, and several of those open pull requests are fixes for real defects rather than features.

## What I am doing meanwhile, so that nobody feels hurried

I have four pull requests open here (#509, #512, #513, #514). Rather than wait on them, I am going to build against a fork of this repository carrying those four patches, so that my own project is not blocked.

I want to be clear about what that is and is not:

- It is not a competing project and I am not inviting anyone to it. It is a private dependency for one downstream user. It carries these four patches, and may carry further fixes as I keep building, but each of those I would open here first, so it stays a fork that is behind upstream rather than one going somewhere else.
- I would rather be on upstream, and I will move back to `main` as soon as there is a release to move to, whether or not my patches are in it. Carrying a fork is a cost I am taking on reluctantly.
- This is not a deadline for anyone. The pull requests can sit as long as they need to. I am saying this precisely so that nobody feels they owe me a merge; I have unblocked myself and the pressure is off.

If any of the four are wrong, I would genuinely rather be told than merged #514 already has a correction on the thread where my first reasoning was wrong, and I would sooner that happened again than ship something a maintainer would not have accepted.

## The question

So, one question, and any answer is better than none: **is the project still maintained, at whatever pace?**

If the answer is "not at the moment", that is a perfectly good answer and I will plan around it. I am not asking for a commitment, a schedule, or a review, just for whatever the honest state of it is.

[...]

PR #519, opened 2026-08-16
server: do not discard the server's own delivery on an inbound PUBLISH carrying the same packet identifier

ifnesi:
# server: do not discard the server's own delivery on an inbound PUBLISH carrying the same packet identifier

> This is the same defect as #514, in the function that pull request held up
> as asking the right question. `processPublish` asks it in one branch and
> then falls through to two operations that do not.

## What happens

A QoS 1 or 2 message the server has outbound to a client is silently
destroyed whenever that client publishes carrying the same packet
identifier. The delivery is then never acknowledged, never resent, and the
send quota it spent is never returned.

Both sides' counters start at 1 on a fresh connection, so a client that
receives and publishes on the same connection collides constantly rather
than occasionally: a device that reports telemetry and accepts commands, a
request/response client, a bridge.

`processPublish` reaches `cl.State.Inflight` — the map of the packets the
**server** has in flight, keyed by the identifiers the **server** assigned —
with the **client's** identifier, twice:

1. the duplicate check deletes whatever it finds (`[MQTT-4.3.2-5]`);
2. for QoS 1 it then `Set`s the outgoing `PUBACK` under the same key and
   `Delete`s it again.

Either one removes the server's delivery. `processPuback` then finds no
entry for that identifier and returns early, so `OnQosComplete` never fires
and `IncreaseSendQuota` is never reached.

## Why it is wrong

Packet Identifiers assigned by the Client and by the Server are independent.
MQTT 5.0 §2.2.1:

> The Client and Server assign Packet Identifiers independently of each
> other. As a result, Client-Server pairs can participate in concurrent
> message exchanges using the same Packet Identifiers.

The non-normative comment beneath it describes precisely this case: a Client
sending a PUBLISH carrying `0x1234` while receiving a *different* PUBLISH
carrying `0x1234` from its Server, before either is acknowledged.

`cl.State.Inflight` holds entries from **both** spaces — the same table
#514 gives:

| entry type | whose identifier | matching an inbound PUBLISH against it |
|---|---|---|
| `Publish` | the server's outbound message | wrong — a different space |
| `Pubrec` | the client's own QoS 2 publish, mid-`PUBREL` exchange | correct — genuinely in use |

The `Pubrec` branch above already makes that distinction correctly. What
follows it does not.

The outcome is a QoS 1 delivery the server accepted and then neither
delivered nor retried, which is the at-least-once guarantee of §4.3.2.

## The change

Two narrowings in `processPublish`, no removals:

**The duplicate check now acts only on a `Pubrec` entry**, which is the one
row of the table above that shares an identifier space with an inbound
publish.

**The `PUBACK` no longer takes an entry in `Inflight`.** It is written
immediately and waits for nothing, so it needs none — the `Set` and the
`Delete` were undone within the same function. A `PUBREC` does wait for a
`PUBREL` and still takes one, under the client's identifier, which is the
space the rest of that flow uses.

One consequence worth flagging rather than burying: `OnQosPublish` and
`OnQosComplete` no longer fire for the `PUBACK` of an inbound QoS 1 publish.
Nothing in the tree depends on it and the storage hooks pass unchanged, but
it is visible to hook authors. Firing `OnQosPublish` for a packet that was
never set in flight seemed the worse of the two, and I am happy to take
direction.

## Tests

`TestServerProcessPacketPublishDoesNotDiscardServerDelivery` is new. It puts
a QoS 1 delivery in flight under identifier 7, has the client publish at
QoS 1 under its own identifier 7, and requires the delivery to survive with
its topic intact. It fails on `main` and passes with the change.

`TestServerProcessPacketPublishQos1PacketIDInUse` seeds an outbound
`Publish` — the first row of the table — and required `Info.Inflight` to
fall to 0, which is the behaviour above. One assertion changes to 1, with a
comment saying why. Its counterpart
`TestServerProcessPacketPublishQos2PacketIDInUse` seeds a `Pubrec`, the
second row, and is untouched — which is the distinction the change turns on.

`go test ./...` and `go test . -race` both pass.

## How it was found

A work queue built on mochi, where a worker answers each job with a QoS 1
publish while the server delivers the next at QoS 1, so the collision is
systematic rather than occasional. A worker declaring Receive Maximum 20
against a backlog of 300 jobs stopped at 22 and was never offered another,
each answer having destroyed one delivery's accounting until the send quota
reached zero and stayed there. With this change the same worker drains all
300.

As on #515, and better said first than discovered later: what is built on
top is itself an MQTT broker.


ifnesi:
One thing worth saying before CI does, so it is not read as this change:
`TestServerProcessPublishACLCheckDeny` fails intermittently under `-race`
on `main` as it stands, at roughly one run in five.

Measured by interleaving whole-package runs of `go test . -count=1 -race`
between the two, fifteen rounds each on one machine: **2 of 15 on `main`,
4 of 15 with this change.** Two against four out of fifteen is noise
around that rate rather than a difference, and the number that matters is
that untouched `main` fails it at all.

It looks like the assertion rather than the server. The subtest ends with

```go
buf, err := io.ReadAll(r)
...
require.Equal(t, tx.expectDisconnect, cl.Closed())
```

and `io.ReadAll` returns when the goroutine closes the write end of the
pipe, which is not ordered against the client being marked closed — so
whichever of the two wins decides the run. `v4_QOS1` and `v4_QOS2` both
show it, whichever gets there first.

Happy to open a separate pull request for it if that would be useful; it
seemed better kept out of this one.

PR #531, opened 2026-08-23
clients: bound the write that holds the client lock

ifnesi:
# clients: bound the write that holds the client lock

## What happens

A client that stops reading its socket, with the connection still open,
holds its own client lock for as long as it stays connected — and every
other goroutine that needs that lock waits behind it.

`WritePacket` takes `cl.Lock()` and then writes to the connection inside
it. A full socket buffer blocks that write with the lock held.
`publishToClient` needs the same lock for a packet identifier:

```go
i, err := cl.NextPacketID() // [MQTT-4.3.2-1] [MQTT-4.3.3-1]
```

and that is *before* the `select` that puts the packet on the outbound
queue or, past `MaximumClientWritesPending`, discards it and calls
`OnPublishDropped`.

**So the shedding this server already implements is unreachable at QoS 1
and 2.** The queue bound, `MessagesDropped` and `OnPublishDropped` are all
behind a lock the slow client is holding. At QoS 0 no identifier is
needed, the `select` runs, and the drop happens exactly as documented —
which is why this is easy to miss.

## Measured

One subscriber that completed its SUBSCRIBE and then stopped reading, and
a publisher flooding a topic that matches it:

| deaf subscriber at | publisher | second publisher, unrelated topic |
|---|---|---|
| QoS 0 | 6,000 acknowledged in 2.3s, drops counted | acknowledged |
| QoS 1 | **stuck at #659 after 8.2s**, nothing dropped | **never acknowledged** |

The second publisher matters: it is blocked on the same lock from inside
`processPublish`, on its own read goroutine, so everything it sends
afterwards waits too — including to topics the deaf subscriber has no
interest in.

A goroutine dump from the stuck server names it:

```
goroutine [sync.Mutex.Lock]:
  mochi.(*Client).NextPacketID       clients.go:277   ← cl.Lock()
  mochi.(*Server).publishToClient    server.go:1179
  mochi.(*Server).publishToSubscribers
  mochi.(*Server).processPublish

goroutine [IO wait]:
  net.(*conn).Write
  mochi.(*Client).WritePacket.func1  clients.go:625   ← under cl.Lock()
  mochi.(*Client).WriteLoop
```

## The change

`Options.ClientNetWriteTimeout` bounds one write.

**Zero is the default and changes nothing for anyone.** Every existing
deployment behaves exactly as it does today unless it sets the option.

Set, a write that does not finish in time fails with a network timeout,
the lock is released, and the caller decides what to do with a connection
that now has part of a packet on it.

**The deadline is armed inside the same critical section as the write** and
cleared before the lock is released. A connection deadline belongs to the
connection rather than to a write, so arming it outside the lock would let
one goroutine clear the deadline another was relying on — the ordering is
the point, not just the presence of a deadline.

## The second half: a timeout stops the client

Bounding the write is necessary and **not sufficient**, which only showed
up when it was run against a real backlog.

`WriteLoop` logs a failed write and takes the next packet. So a client that
has stopped reading is written to once per timeout for every packet still
queued, and each attempt takes the client lock again. Measured with 4,000
queued deliveries: the writes did time out, once every five seconds as
configured, and a goroutine on an unrelated topic was **still** starved of
that lock twenty seconds later.

It is also wrong on its own terms. A write that ran out of time has put
part of a packet on the wire; the next packet written lands in the middle
of the last one, so a client that ever read again would get a corrupted
stream.

So a timeout stops the client. Other write errors keep the behaviour they
have, `TODO` and all — this is the one case where continuing is knowably
worse than stopping.

**A timeout is recognised by asking `net.Error`,** not by matching
`os.ErrDeadlineExceeded`. The two agree for a plain socket write and do not
for every listener — see below, which is where that cost something.

## The third half: the deadline has to reach the write

`ClientNetWriteTimeout` arms the deadline on `cl.Net.Conn`. For the TCP and
Unix listeners that is the socket, and it works. For the websocket listener
it did nothing at all, silently:

- `wsConn` embeds `net.Conn`, so `SetWriteDeadline` was promoted from the
  socket and looked like the right method;
- gorilla calls the socket's `SetWriteDeadline` with its own stored
  deadline immediately before every frame it writes, so the deadline set
  above is overwritten — with the zero value, meaning none — before the
  write it was meant to bound.

A client that opens a websocket, subscribes and stops reading therefore
held the client lock exactly as before, with the option set and accepted.
Measured on a broker configured to time out after 300ms: a publisher stuck
at its 1,014th message.

`wsConn` now has a `SetWriteDeadline` of its own, delegating to the
websocket connection, which is where gorilla reads the deadline it applies.

The same listener is why the timeout check is `net.Error`: gorilla's
buffered `writev` yields `writev tcp ...: i/o timeout`, which is a timeout
by every definition that matters and does not match
`os.ErrDeadlineExceeded`. With the sentinel, `WriteLoop` went on to the
next packet and never stopped the client — one transport covered and
another not, from a check that reads as correct.

One more consequence: the deadline is now cleared on the connection it was
armed on, held in a local, rather than on whatever `cl.Net.Conn` names by
the time the write returns.

## The fourth half: the client can switch the deadline off

`refreshDeadline` calls `SetDeadline`, which sets the read *and* write
deadlines, and it runs at the top of every pass of `Read`. So a client that
sends anything needing no reply — a QoS 0 PUBLISH, a PUBACK, an UNSUBSCRIBE
— takes the loop round again and clears the deadline armed for a write
already blocked on that client's socket. With keepalive 0 the refresh is the
zero time, which clears it; with keepalive 60 it is 90 seconds out, and the
next packet pushes it out again.

Everything above therefore held only for a client that stopped reading *and*
stopped sending. One that stops reading and goes on publishing holds the
client lock exactly as it did before this option existed. Measured on a
broker built on this server, deaf QoS 1 subscriber publishing to an
unrelated topic every 200 ms:

| | |
|---|---|
| keepalive 0 | second subscriber received 0 of 5 in 40.1 s, 0 disconnections |
| keepalive 60 | second subscriber received 0 of 5 in 40.1 s, 0 disconnections |

A PINGREQ does not do it, which is why the tests above passed: the PINGRESP
queues behind the same lock, nothing more is read, and the loop never comes
round.

`refreshDeadline` now sets the read deadline only. [MQTT-3.1.2-22] is a rule
about packets *received* — "the Server MUST close the Network Connection if
it does not receive a PUBLISH, PUBACK ... within one and a half times the
Keep Alive time period" — so a read deadline is what it describes. Setting a
write deadline alongside it was harmless while nothing else set one, and it
is not this option's to keep.

## What this does not do

It does not change the outbound queue, `MaximumClientWritesPending`, or any
behaviour at QoS 0. And it is one write, not a rate: a client that reads
slowly but steadily is unaffected, because a write only waits at all once
the kernel's own buffer is full.

## Tests

Four in `clients_test.go`, three of them using `net.Pipe`, which is unbuffered — a
write to it blocks until the other side reads, and nothing reads, which is
what a full socket buffer is:

- a client that never reads returns a timeout **and leaves the lock free**,
  which is the half that matters;
- the default of zero still blocks until a reader arrives and then
  completes normally;
- `WriteLoop` against a pipe nobody reads closes the client, with that
  timeout as its stop cause.

A fourth in `clients_test.go` drives the read loop with packets arriving
while a write is blocked, and asserts the write still times out. With
`SetDeadline` restored it fails in two seconds.

One in `listeners/websocket_test.go`, driving a real gorilla client that
connects and never reads: the server-side write ends. With the deadline set
on the socket instead, it hangs for the full ten seconds the test allows.

Run against the unpatched tree, the first fails:

```
--- FAIL: TestClientWritePacketTimesOutRatherThanHoldingTheLock (2.01s)
    WritePacket to a client that never reads did not return: it is holding
    the client lock, and every other goroutine that needs it —
    publishToClient taking a packet identifier for a QoS 1 delivery —
    waits behind it
```

`go test ./... -race`, `go vet ./...` and `gofmt -l .` are clean on this
branch, over three consecutive runs.


ifnesi:
Pushed a third commit and updated the description above.

The option worked on the TCP and Unix listeners and did nothing on the websocket one. `wsConn` embeds `net.Conn`, so `SetWriteDeadline` was promoted from the socket — and gorilla overwrites the socket's deadline with its own, the zero value, immediately before every frame. A websocket client that subscribed and stopped reading held the client lock with the option set and accepted; measured with a 300ms timeout, a publisher stuck at message 1,014.

`wsConn` now has its own `SetWriteDeadline`, delegating to the websocket connection. The same listener is why `WriteLoop` now asks `net.Error` whether the write timed out instead of matching `os.ErrDeadlineExceeded`: gorilla's buffered `writev` yields `writev tcp ...: i/o timeout`, which is not that sentinel, so the loop went on to the next packet and never stopped the client.

There is a listener test driving a real gorilla client that never reads; it hangs for the full ten seconds against the socket deadline. The deferred clear also now holds the connection in a local rather than re-reading `cl.Net.Conn`, so it clears the deadline on the connection it armed.

ifnesi:
Fourth commit, and it is the one that mattered most: everything above held only for a client that stopped reading **and** stopped sending.

`refreshDeadline` calls `SetDeadline` — read and write together — at the top of every pass of `Read`. So any packet needing no reply (a QoS 0 PUBLISH, a PUBACK, an UNSUBSCRIBE) takes the loop round again and clears the deadline armed for a write already blocked on that socket. Keepalive 0 clears it outright; keepalive 60 pushes it 90 seconds out and the next packet pushes it again.

Measured on a broker built on this server, with the deaf subscriber publishing to an unrelated topic every 200ms: a second subscriber received nothing for 40 seconds, no disconnections, at both keepalive settings. A PINGREQ does not do it — the PINGRESP queues behind the same lock and nothing more is read — which is why every test here passed: they all used a client that went silent.

`refreshDeadline` now sets the read deadline only. [MQTT-3.1.2-22] is about packets *received*, so that is what it describes; the write deadline alongside it was harmless until something else set one. New test drives the read loop with packets arriving while a write is blocked.

PR #537, opened 2026-09-01
server_test: join the goroutine before asserting on the disconnect

ifnesi:
TestServerProcessPublishACLCheckDeny is intermittent on `main`, and it is the failing check on several open PRs — including ones that predate anything touching that path.

| | failures in 120 runs, `GOMAXPROCS=2`, go1.21.13 |
|---|---|
| `main` | **3** |
| this branch | **0** |

go1.21.13 because that is what CI pins. The rate is higher on newer toolchains — 5 and 3 per 40 on go1.25.

The test reads the connection until it closes, asserts the client was disconnected, and only then joins the goroutine doing the disconnecting. Reading until close does not mean the writer has finished — the disconnect under test is one of the things that closes the pipe — so `cl.Closed()` is read while `Stop` is still running. A fast machine usually wins that race; a two-core runner does not.

Joining before the assertions also removes the data race the CI logs show alongside the failure: a failed assertion calls `FailNow`, which skips a `Wait` placed after it, leaving a goroutine reading the shared range variable that the next iteration writes. `go.mod` declares `go 1.21`, so that is one variable for the whole loop.

One statement moved. No production code.


ifnesi:
Duplicate of #530, which does the same thing and one more — it moves the `wg.Wait()` above the assertions *and* captures the range variable. #530 has been open since 23 August and is green; please use that one. Apologies for the noise.

Issue #424, opened 2024-08-01
Function attachClient sometimes does not call Done() on ClientsWg.


NiklasCi:
I using this awesome mqtt broker to test mqtt client implementations. 
I have alot of small tests where i start the broker, publish a message and catch it with a hook.

I notice that sometimes the `server.Close` function indefinitely stuck because it waits for `ClientsWg` turning zero.
I checked where in the code `Add(1)` and `Done()` is called and added some logs above like this:

```go
// attachClient validates an incoming client connection and if viable, attaches the client
// to the server, performs session housekeeping, and reads incoming packets.
func (s *Server) attachClient(cl *Client, listener string) error {
	defer func() {
		s.Listeners.ClientsWg.Done()
		fmt.Println("DONE")
	}()
	fmt.Println("ADD")
	s.Listeners.ClientsWg.Add(1)
	
	// ...
}
```

I found out that sometimes `Add()` is called but not `Done()` which leads to stucking `server.Close()` call.
There is some race condition that needs to be investigated how that can happen.

Here my logs:
```
time=2024-08-01T14:13:02.464+02:00 level=INFO msg="added hook" hook=allow-all-auth
time=2024-08-01T14:13:02.464+02:00 level=INFO msg="added hook" hook=events-test
time=2024-08-01T14:13:02.464+02:00 level=INFO msg=initialised hook=events-test
time=2024-08-01T14:13:02.464+02:00 level=INFO msg="attached listener" id=04337e06-3f96-4f74-bb4d-286191bb5adb protocol=tcp address=127.0.0.1:1883
time=2024-08-01T14:13:02.464+02:00 level=INFO msg="mochi mqtt starting" version=2.6.5
time=2024-08-01T14:13:02.464+02:00 level=INFO msg="mochi mqtt server started"
ADD
ADD
DONE
time=2024-08-01T14:13:02.965+02:00 level=WARN msg="" listener=04337e06-3f96-4f74-bb4d-286191bb5adb error="read tcp 127.0.0.1:1883->127.0.0.1:35602: use of closed network connection"
ADD
DONE
time=2024-08-01T14:13:02.966+02:00 level=WARN msg="" listener=04337e06-3f96-4f74-bb4d-286191bb5adb error="read tcp 127.0.0.1:1883->127.0.0.1:35616: use of closed network connection"
ADD
DONE
[...]

PR #540, opened 2026-09-04
Fix two Close()-vs-connect/disconnect deadlock/panic bugs

### LINKED ISSUE #488: Deadlock in `clients.go` due to recursive read-locking

quarksus:
## Summary

Found while stress-testing an embedding project's own mTLS work under heavy concurrent load: `go test -race ./...` genuinely deadlocked and hit the test timeout — not a flake, a real stuck goroutine, reproducing only under CPU contention (running many tests together) and never in isolation. Root-caused to two independent bugs, both confirmed with a minimal, consumer-code-free reproduction harness (a real `mqtt.Server`, real TCP connections, no application code involved) before deciding on fixes.

### Bug 1 — permanent deadlock (fixes #488)

`Clients.GetByListener` holds `cl.RLock()`, then calls `cl.Len()`, which acquires the *same* read lock again. `sync.RWMutex`'s own docs call this out as unsafe: once a concurrent `Lock()` (`Clients.Add`/`Delete`, called on every connect/disconnect) is pending, further `RLock()` calls block until the writer is served — but the writer can never be served because it's waiting on the very read lock stuck behind it. `GetByListener` is called from every `Server.Close()`, so any `Close()` racing a live connect/disconnect can trigger it. This is the exact bug already reported in #488, with the identical fix independently arrived at (also proposed in #489) — read `len(cl.internal)` directly, since the lock is already held.

### Bug 2 — WaitGroup ordering race, hang or panic (same root cause as #424)

`TCP.Serve`'s accept loop spawns each connection's handler as an untracked, fire-and-forget goroutine. `Listeners.ClientsWg.Add(1)` only happens deep inside that goroutine (in `Server.attachClient`), with no ordering guarantee relative to a concurrent `Close()`'s `Listeners.ClientsWg.Wait()` — textbook `WaitGroup` misuse per the `sync` package's own docs ("calls with a positive delta that start when the counter is zero must happen before a `Wait` call"). This is the same underlying issue as #424, which was closed without ever being fixed (the original reporter's own test just stopped reproducing it). Independently reproduced here as both `Close()` hanging indefinitely and, under `-race`, a `sync: WaitGroup is reused before previous Wait has returned` panic.

**Fix:** track in-flight `establish()` calls with the listener's own `WaitGroup`, `Add()`'d synchronously on the accepting goroutine *before* the handler goroutine is spawned, with `Close()` waiting on it before returning. By the time `TCP.Close()` returns, every `establish()` call it ever started — and so every `ClientsWg.Add`/`Done` pairing it made — has already fully completed, sidestepping the ordering hazard entirely. Also closes a minor pre-existing fd leak: a connection accepted in the instant `Close` was already tearing down the listener was previously left open with nothing ever closing it.

### Why these are two commits, not one

They're independent root causes, but they share the exact same trigger condition (`Close()` racing connect/disconnect), so an end-to-end regression test exercising `Server.Close()` for real needs *both* fixed to pass reliably — that's reflected in `TestServerCloseWaitsForInFlightConnections`.

## Tests

- `listeners/tcp_test.go`: `TestTCPCloseWaitsForInFlightEstablish` — a deterministic unit test that `Close()` actually blocks on an in-flight `establish()` call.
- `server_test.go`: `TestServerCloseWaitsForInFlightConnections` — an end-to-end regression test racing real `Close()` against real connection churn through a real `Server`. Confirmed to fail (hang, or a `WaitGroup` panic under `-race`) on the pre-fix code within ~150 iterations across several runs; passes reliably (200 iterations × several runs) after both fixes.

**One note for reviewers**: that same `Close()`-vs-`Serve()` scenario can occasionally also surface a *separate*, pre-existing data race under `-race`, unrelated to either fix here — confirmed present on unpatched v2.7.9 too. I didn't want to scope-creep this PR trying to fix it, so I've filed it separately as #541; there's a note in the test itself pointing here so it isn't mistaken for a regression.

## Verification

Both fixes verified against a minimal reproduction harness (no consumer code, just `mqtt.New` + a real TCP listener + real connect/disconnect churn racing `Close()`): unpatched, bug 1 triggered in 33–67 iterations and bug 2 within ~150; patched, 3000+ iterations with zero recurrence of either, plus four *concurrent* full `go test -race ./...` suite runs of the embedding project (the exact contention pattern that originally surfaced this in its CI) — all clean.

🤖 Part of this description was drafted with [Claude Code](https://claude.com/claude-code), based on my own investigation and reproduction of both issues.


listeners/tcp.go
```diff
@@ -81,29 +90,43 @@ func (l *TCP) Serve(establish EstablishFn) {
...
+			// establish() this listener ever started really has finished.
+			l.connWg.Add(1)
 			go func() {
-				err = establish(l.id, conn)
+				defer l.connWg.Done()
+				err := establish(l.id, conn)
 				if err != nil {
 					l.log.Warn("", "error", err)
 				}
 			}()
```
Nocccer, line 105: What if we use a errgroup instead a `sync.Waitgroup`. We can react to errors if needed aswell as can use `errgroup.Go()` to avoid using `Add` and `Done`.

Also if the go version is new enough you could use `l.connWg.Go` aswell to avoid `Add` and `Done`

